PR #17, opened 2019-06-23
Preserve network metadata in save() and load()

ivan-pi:
Hello Milan,

The save and load functions should now also load the metadata. I was not sure what to do about the first layer, is it always supposed to have a sigmoid function? I made the function simply save and load all layers.

I took the freedom to change the `set_activation` function of the `layer_type` to an elemental subroutine and added a second thin wrapper in the `network_type` class. This way, different functions can be set for each individual layer using the syntax:
```Fortran
call net % set_activation([character(len=10) :: 'sigmoid', 'tanh', 'gaussian']) ! 3 layer network
```
The same could be done for the network constructor.

The `test_network_save` test has been modified to test also the activation functions are matching (by procedure pointer association).

This pull request will make old saved network files broken, and will need to be manually edited to include the missing metadata.

### DIFF
--- a/src/lib/mod_layer.f90
+++ b/src/lib/mod_layer.f90
@@ -18,6 +18,7 @@ module mod_layer
     real(rk), allocatable :: z(:) ! arg. to activation function
     procedure(activation_function), pointer, nopass :: activation => null()
     procedure(activation_function), pointer, nopass :: activation_prime => null()
+    character(len=:), allocatable :: activation_str ! activation character string
   contains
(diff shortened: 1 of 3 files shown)


milancurcic:
> I was not sure what to do about the first layer, is it always supposed to have a sigmoid function?

For feed-forward networks (currently the only kind supported), the first layer is only an input layer so the activation function is never used. Other kinds of networks may use it, so let's keep this the way you made it for the time being.

PR #33, opened 2021-07-19
Fix typo in fpm build instruction 

milancurcic:
Thanks @zoziha!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Dependencies:
 #### Building in serial mode
 
 ```
-fpm build --flag "-cpp -O3 -ffast-math fcoarrays=single"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,7 +46,7 @@ Dependencies:
 #### Building in serial mode
 
 ```
-fpm build --flag "-cpp -O3 -ffast-math fcoarrays=single"
+fpm build --flag "-cpp -O3 -ffast-math fcoarray=single"
```
zoziha, line 49: Hmm, is there still lack of a dash here, before `fcoarray=single`, like this:
```bash
-fcoarray=single    # "-"
```

PR #52, opened 2022-04-12
Delete CAF C preprocessor macro

rouson:
This PR includes all of the commits from PR #51 plus one additional commit that removes the `CAF` C preprocessor macro.  Because the macro appears only in procedure definitions and never in interfaces, it would be impractical to create this PR without including the commits from PR #51 because the latter PR moves all procedure definitions to new files.


milancurcic (approved):
Looks good, thank you!

Issue #54, opened 2022-04-15
Download data files at runtime or in an installation script


rouson:
I just uploaded [mnist.tar.gz](https://github.com/modern-fortran/neural-fortran/files/8498876/mnist.tar.gz) to GitHub file storage in way that is persistent without being part of the `git` repository.  I recommend against storing binary files in a repository.  Over time, they increase download times and because `git` can't do useful diffs on binary files, every change to the file means a completely new version must be stored.  What's worse, `git rm` doesn't take any of the committed versions of the file out of the commit history so the only way to fix the problem is to rewrite history, which essentially means that everyone who has a local copy of the repository will need to do a fresh clone and if having everyone do a fresh clone is not practical, then it's probably necessary to set up a new repository.

I recommend either

1. Use Fortran's intrinsic `execute_commnand_line()` subroutine to download and uncompress the file from the above location at runtime if it's missing or
2. Provide a short script that users can run to handle basic set up tasks (including downloading data files) and then build and test the repository's software.

On the OpenCoarrays project, I found that our downloads went up by a factor of 2-3 soon after I wrote an installation script. More recently, I settled on a much simpler approach to writing an installer for [Caffeine](https://go.lbl.gov/caffeine) that is an order of magnitude smaller than the OpenCoarrays installer, more robust, and much more maintainable.  **I'll be glad to adapt Caffeine's `install.sh` script to neural-fortran if you like.**  When someone can get a package built and tested by typing nothing more than `./install.sh`,  it saves a lot of time over reading build/test instructions, no matter how simple those instructions are.

milancurcic:
I agree and sounds good. I didn't know that GitHub offered file storage, thanks for uploading it.

I'm fine with either of the two approaches. Option 1 seems more attractive, because option 2 would be doing much of what fpm and CMake are already doing. So install.sh would be downloading the data automatically, but for fpm we'd still need separate instructions for manual download of the data. Alternative would be to have a script to download the data.

Option 1 would also have a nice UX that Keras has with its toy [datasets](https://keras.io/api/datasets/). It works just like you wrote--the first time you invoke the function to load the dataset, it's downloaded to a conventional directory, and loaded from disk on any subsequent call. Even if you decide to add the script, we'll be looking at this option long-term.

milancurcic:
Another good reason to do this with the dataset is that with the addition of CNNs and hopefully other architectures (RNNs, etc.), I'd want to add several other toy datasets.

[...]

PR #55, opened 2022-04-17
Download mnist.tar.gz if it is missing

rouson:
This PR resolves issue #54 and contains all of the commits from PR #51 because it modifies submodules that won't exist on the main branch until #51 gets merged.


rouson:
After this commit gets merged, it would be a good idea to use a git repo surgery tool to remove all copies of mnist.tar.gz and any other binary files from the repositories history.  @everythingfunctional recently did this on another project and might have advice about what tool to use.  After removing all binary files from the commit history, anyone who has a local copy of this repository will need to obtain a fresh clone -- especially if they will ever push commits to this repository or submit a pull request from a fork created before the binary files were removed.

milancurcic:
Thanks, Damian. I moved the subroutine to the MNIST submodule because mod_io provides more general functions to read binary files. Let me know if it looks good. I'll merge tomorrow if no objections.

Eventually we should generalize and move the downloader function into its own module as there will be more datasets to download.

milancurcic (approved):

rouson:
@milancurcic looks good to me.  Thanks for approving the PR.

rouson (commented):
LGTM

rouson:
@milancurcic let me know if you would like any refinements to this PR before it's merged.  Examples:

1. The code could check whether mnist.tar.gz is present, and if so, not download it.
2. The code could prompt the user for permission to download the file before doing so.

These are the sorts of things I would do if I were writing a shell script to download the code, but I'm on the fence as to whether these would be improvements for several reasons:

1. Each `execute_command_line` call requires assumptions about what type of OS shell is being used (must be Unix-like) what's available on the system and what commands are available in that shell (e.g., `curl` and `ls`).
2. `execute_command_line` requires verbose arguments and verbose error handling, some of which might not even work depending on the system behavior (the error-handling code in this PR seemed to end up being a no-op in macOS -- I haven't tested it in Linux).
3. Even if the file is present, it might be from a failed or incomplete download so presence of the file doesn't necessary guarantee the integrity of the file, which is why I defaulted to just doing the download if the uncompressed files are missing.

milancurcic:
No, let's do any refinements in separate PRs. I prefer small PRs.

We already check for presence of one .dat file to determine whether to download or not. This assumes that all files are present and correctly downloaded. We can improve this, but in a separate PR.

As you work with me you'll find that I'm all about "good enough" and "mostly working" solutions, and never about bullet-proof or perfect solutions. This allows me to find out what's a real problem and what isn't, instead of guessing.

rouson:
@milancurcic We are very much aligned: I frequently advocate for getting working solutions out as early as possible and then making refinements based on the feedback of users or other developers.

PR #71, opened 2022-05-18
Add UML diagrams

rouson:
This PR adds UML documentation in the doc/ subdirectory, including

1. A README.md describing and depicting the diagrams.
2. A UML class diagram detailing the derived type relationships.
3. A UML class diagram documenting the developer API comprised of all public derived types, type-bound procedures, and generic interfaces, including user-defined structure constructors.

To Do (in a future PR)
---------------------
Add a User API diagram depicting only the derived types, type-bound procedures, and constructors intended for use in applications.


rouson:
@milancurcic these diagrams can be especially helpful for a high-level picture of what's in neural-fortran with some information that doesn't appear in FORD graphics (e.g., the different types of class relationships such, including type extension and aggregation/components).  As one example of the utility, I naively misinterpreted `base_layer` as referring to the first of several layers in a deep neural network when in fact "base" refers to a class relationship rather than an internal structure in the network.  Having a readily accessible UML diagram clears up such misinterpretations very quickly.  A picture is worth a thousand words. :)

rouson:
@milancurcic if you can tell me what derived types are intended to be user-facing versus which ones are strictly for internal use, then I'll contribute one more pull request with just those to describe the "user API."  This too is really helpful for a quick introduction to neural-fortran for newcomers.

rouson:
@milancurcic I would also be glad to contribute some UML sequence diagrams that would depict the expected runtime behavior of neural-fortran, including the sequence of type-bound procedure invocations in a common use case, the objects involved, and the lifetimes of those objects.

milancurcic:
Thanks @rouson. I'd like to have diagrams like this.

What's the advantage of these diagrams over the ones generated by FORD and graphviz? I'm reluctant to add yet another developer documentation over the one we already have.

As the API will change and grow in the near future, I suggest not committing the UML sources to the repo, but merely providing a script or instruction how to generate them. Otherwise, the diagrams will quickly fall out of date.

The user API is/will be what the [nf](https://github.com/modern-fortran/neural-fortran/blob/main/src/nf.f90) module provides. I say "will be" because I just realized that the MNIST example uses an optimizer and MNIST loaders directly from their own modules. I think it will make sense, and be convenient, to organize the public API by themes, e.g. `nf_datasets`, `nf_optimiziers`, etc. I'm not decided yet on that and open to suggestions.

It's unlikely that the users will want to fiddle with `base_layer` or anything that extends it. The best way to explore the API is to start from the examples and work your way down.

rouson:
@milancurcic everything you mentioned is already addressed in the PR, including examples of ways in which FORD and UML serve  distinct purposes.  To wit,

1. This PR commits scripts, not diagrams, and the new doc/README.md explains how to generate the diagrams.
2. The new doc/README.md demonstrates an aspect of a derived-type relationship that UML shows clearly and succinctly but FORD diagrams do not.   My comment above explains a source of confusion that drawing the UML diagrams cleared up for me regarding one aspect of the project nomenclature.  I'm offering the PR in hopes that it could save other newcomers time because they'll discover such things immediately upon inspection of the diagram without having to open file after file to figure it out.   FORD diagrams don't describe the nature of derived-type relationship, e.g., type extension versus aggregation and one-to-one versus one-to-many.  UML can show all of this and a lot more that I'm not even showing yet but will eventually (e.g., correctness constraints can be expressed formally in UML).   By contrast, FORD gathers information about argument attributes (e.g., `intent`).  
3. PlantUML scripts are both lightweight (some as short as 20 lines) and their syntax resembles the UML diagram that they generates, which makes it extremely easy to keep them up-to-date as the code evolves (seconds or minutes of time to add a new class or class relationship).
4. The README.md has separate sections for "Developer API" and "User API".   The latter should have only small subset of what is in the former and I can't see anywhere else that the project tells me what that subset is.  I'm glad to take the time to discover it myself because I want to l earn neural-fortran and along the way learn UML.  Alternatively, you can tell me and I'll handle creating the related scripts.  My goal is to insure that everything I learn in this process shortens the learning curve for future users or contributors.

milancurcic (approved):
I fixed the typos and removed the redundant info. I'll let you finish up the User API diagram.

milancurcic:
@rouson, I just noticed that you wanted to do the User API UML in another PR. I agree. If this is ready from your end, let's merge it.

PR #74, opened 2022-05-19
fix #72 - use dir tree to expose user API

rouson:

PR #77, opened 2022-05-20
Fix CMake build for the new src directory structure

milancurcic:
We forgot to edit the CMakeLists for the new src directory structure in #74. This fixes that.


rouson:
Sorry I missed it.

Issue #43, opened 2022-04-01
Keras HDF5 reader


milancurcic:
Read Keras models from h5 files.

Needs #42.

milancurcic:
Minimal reader (input and dense layers) is implemented. Let's track expanding functionality in specific issues.

PR #79, opened 2022-06-06
Support for loading Keras models

milancurcic:
This PR introduces support for reading simple Keras models from HDF5:

* New `network` constructor that creates a network from an HDF5 file
* Only input and dense layers are supported for now
* Only loading for inference is supported from now; i.e. optimizer data is ignored
* New example (`mnist_from_keras`) that demonstrates loading a pre-trained model and checking the accuracy of the output
* Introduces dependency on HDF5, h5fortran, and json-fortran
* More general `download_and_unpack` subroutine to download tar-balled data
* Placed `nf_io_*` modules under src/nf/io.
* Modernized CMake rules

TODO:

* [x] Add dependency rules to CMakeLists.txt (thanks @scivision)
* [x] Test with Intel Fortran

In support of #43.


milancurcic:
@scivision, say I have an Intel build of HDF5 in `/opt/hdf5-intel`, with library and module files in lib/ and include/ relative to that path. 

How do I pass this information to CMake so that it finds it?

I tried

```
FC=ifort cmake .. -DSERIAL=1 -DCMAKE_LIBRARY_PATH=/opt/hdf5-intel
```

which doesn't seem to do it.

I'd like to add this instruction to the README for people who want to run with Intel.

milancurcic:
All good, I figured it out--it expects `HDF5_ROOT` to be set.

PR #83, opened 2022-06-10
fixed a memory leak when reading the JSON

### LINKED ISSUE #82: Memory leak in get_keras_h5_layers
In `get_keras_h5_layers`, this code:

```fortran
    call json % parse(model_config_json, model_config_string)
```

allocates the local `model_config_json` pointer, but it is not destroyed in this routine, so that's a memory leak. You should add a:

```fortran
   call json % destroy (model_config_json)
```

At the end to free the memory. The other `json_value` pointers all look like they are just pointing to data inside this structure, so you don't have to worry about those.

FYI: the higher-level `json_file` class is a little safer (but less flexible) to use since it will destroy the pointer when it goes out of scope (it has a finalizer). What you have here is fine though (you just need the `destroy` call at the end).

jacobwilliams:
Fixes #82

### DIFF
--- a/src/nf/nf_keras_submodule.f90
+++ b/src/nf/nf_keras_submodule.f90
@@ -65,6 +65,9 @@ contains
 
     end do layers
 
+    ! free the memory:
+    call json % destroy(model_config_json)
+
   end function get_keras_h5_layers
 
 end submodule nf_keras_submodule


milancurcic:
Thanks!

And just so I'm sure I understand: The reason we don't need to destroy other `json_value` pointers, in this case, is because they are all pointing to child nodes of `model_config_json`, so by destroying it, we free the whole structure including the child nodes, yes?

jacobwilliams:
Yep, that is correct. The `destroy` call deallocates the entire tree structure. 

The other variables are pointers to the existing structure. When you destroy the structure they are now dangling pointers, but since they are just local variables it doesn't matter.

PR #106, opened 2022-12-02
getting / setting network parameters

jvo203:
* fixed CMakeList.txt
* added ```get_num_params()```
* added ```get_parameters()```
* added ```set_parameters()```
* created an example test program: ```example/params.f90```
* re-formatted the changed Fortran files in vscode under Linux instead of macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,9 @@ include(cmake/json.cmake)
 # library to archive (libneural.a)
 add_library(neural
   src/nf.f90
-  src/nf/nf_activation.f90
+  src/nf/nf_activation_1d.f90
+  src/nf/nf_activation_3d.f90
   src/nf/nf_base_layer.f90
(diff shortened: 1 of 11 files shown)


milancurcic:
Thanks for this. Unfortunately, the spurious indent changes remain. Whatever VS Code plugin is doing this, I don't recommend using it (try https://marketplace.visualstudio.com/items?itemName=fortran-lang.linter-gfortran instead). Alternatively, there should be a way to change the indent level in your plugin to 2, which I expect will revert the spurious indents. It'd be great if you find a way to do it. If not, I can address it when I get back to work on Dec 12.

jvo203:
OK, I'll have another go at formatting. On both Linux and macOS I use the "Modern Fortran" extension from the vscode marketplace when editing .f90 files. But I might also have installed findent and/or fprettify somewhere along the way.

jvo203:
Closing this Pull Request in order to submit a new one with re-formatted Fortran files.

PR #107, opened 2022-12-03
getting / setting network parameters (third-time lucky with tab space formatting)

jvo203:
* fixed CMakeList.txt
* added get_num_params()
* added get_parameters()
* added set_parameters()
* created an example test program: example/params.f90
* re-formatted in vscode on macOS using "Modern Fortran" with the tab indent set to 2 spaces (hopefully)

Regarding the indenting, in the macOS vscode I have already been using the "Modern Fortran" extension as recommended by you. The "fprettify" vscode extension is not installed. Searching the installed extensions for "fortran" only returns "Modern Fortran", no other stuff is installed. In the vscode I have just set "Editor: Tab Size" to 2 and disabled automatic tab size recognition from existing files. Then I touched and saved all the modified .f90 files, triggering automatic re-formatting upon save with the tab size set to 2. Hopefully the files now look more like what you expect them to be.

Please let me know if there are any more settings that might need changing in vscode. If this PR does not fix indenting then it might have to be "over to you" upon your return to work. No worries, this stuff can wait!

### DIFF
--- a/example/params.f90
+++ b/example/params.f90
@@ -48,4 +48,8 @@ program params
    nparam = net % get_num_params()
    print '("#parameters: ", i0)', nparam
 
+   if (allocated(parameters)) then
+      print *,'parameters:', parameters
+   end if
+
 end program params
--- a/example/params.f90
+++ b/example/params.f90
@@ -1,79 +1,79 @@
 program params
-    use nf, only: dense, input, network
-    implicit none
-    type(network) :: net, net2
-    real :: x(1), y(1)
-    real, parameter :: pi = 4*atan(1.)
-    integer, parameter :: num_iterations = 100000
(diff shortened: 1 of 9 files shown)


jvo203:
An improvement: in the top file ```nf_network_submodule.f90```, the subroutine ```get_parameters(self, params)```, it makes sense to add
```
if (allocated(params)) deallocate(params)
```
right before the loop
```
do n = 1, size(self%layers)
```

Some users might repeatedly call ```get_parameters``` (for whatever reason), the fix makes sure the ```params``` array is emptied / deallocated prior to any 2nd-time use. Otherwise parameters would keep getting appended to the end of the existing ```params``` array instead of being written starting at the beginning of ```params```.

PR #111, opened 2022-12-16
Get set network parameters

### LINKED ISSUE #103: extract / set all network parameters via a single 1D real array
Hi, it's just a simple feasibility study but I am considering coupling neural-fortran with a PIKAIA genetic algorithms optimizer (https://github.com/jacobwilliams/pikaia). Is there an easy way to extract all the neural network parameters from all the layers in a single function call, in the form of a long 1D real array? And then use another single subroutine to set all the network parameters by loading them from a 1D real array?

milancurcic:
Opening this in favor of #107 by @jvo203. It should be easier to review. I made some small cosmetic changes.

* [x] Implementation
* [x] Example
* [x] Tests

Closes #103.

### DIFF
--- a/example/get_set_network_params.f90
+++ b/example/get_set_network_params.f90
@@ -49,7 +49,7 @@ program get_set_network_params
   nparam = net % get_num_params()
   print '("get_num_params = ", i0)', nparam
 
-  call net % get_parameters(parameters)
+  call net % get_params(parameters)
 
   if (allocated(parameters)) then
     print '("size(parameters) = ", i0)', size(parameters)
@@ -66,7 +66,7 @@ program get_set_network_params
   call net2 % print_info()
 
   ! copy the parameters from net to net2
-  call net2 % set_parameters(parameters)
+  call net2 % set_params(parameters)
 
   ypred = [(net % predict([xtest(i)]), i=1, test_size)]
   print *, 'Original network test output:'
--- a/example/get_set_network_params.f90
+++ b/example/get_set_network_params.f90
@@ -6,7 +6,8 @@ program get_set_network_params
   real, parameter :: pi = 4*atan(1.)
   integer, parameter :: num_iterations = 100000
   integer, parameter :: test_size = 30
-  real :: xtest(test_size), ytest(test_size), ypred(test_size)
+  real :: xtest(test_size), ytest(test_size)
+  real :: ypred1(test_size), ypred2(test_size)
   integer :: i, n, nparam
   real, allocatable :: parameters(:)
 
@@ -31,13 +32,13 @@ program get_set_network_params
     x = x*2*pi
     y = (sin(x) + 1)/2
 
-    call net%forward(x)
-    call net%backward(y)
-    call net%update(1.)
+    call net % forward(x)
+    call net % backward(y)
+    call net % update(1.)
 
     if (mod(n, 10000) == 0) then
-      ypred = [(net % predict([xtest(i)]), i=1, test_size)]
-      print '(i0,1x,f9.6)', n, sum((ypred - ytest)**2)/size(ypred)
+      ypred1 = [(net % predict([xtest(i)]), i=1, test_size)]
+      print '(i0,1x,f9.6)', n, sum((ypred1 - ytest)**2) / size(ypred1)
     end if
 
   end do
@@ -49,12 +50,12 @@ program get_set_network_params
   nparam = net % get_num_params()
   print '("get_num_params = ", i0)', nparam
 
-  call net % get_params(parameters)
+  parameters = net % get_params()
+  print '("size(parameters) = ", i0)', size(parameters)
+  print *, 'parameters:', parameters
 
-  if (allocated(parameters)) then
-    print '("size(parameters) = ", i0)', size(parameters)
-    print *, 'parameters:', parameters
-  end if
+  print *, 'Now create another network of the same shape and set'
+  print *, 'the parameters from the original network to it.'
 
   net2 = network([ &
     input(1), &
@@ -63,17 +64,11 @@ program get_set_network_params
     dense(1) &
   ])
 
-  call net2 % print_info()
-
   ! copy the parameters from net to net2
   call net2 % set_params(parameters)
 
-  ypred = [(net % predict([xtest(i)]), i=1, test_size)]
-  print *, 'Original network test output:'
-  print *, ypred
-
-  ypred = [(net2 % predict([xtest(i)]), i=1, test_size)]
-  print *, 'Cloned network test output:'
-  print *, ypred
+  ypred1 = [(net % predict([xtest(i)]), i=1, test_size)]
+  ypred2 = [(net2 % predict([xtest(i)]), i=1, test_size)]
+  print *, 'Original and cloned network outputs match:', all(ypred1 == ypred2)
 
 end program get_set_network_params
--- a/example/get_set_network_params.f90
+++ b/example/get_set_network_params.f90
@@ -13,6 +13,10 @@ program get_set_network_params
 
   print '("Getting and setting network parameters")'
   print '(60("="))'
+  print *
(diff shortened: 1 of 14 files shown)


src/nf/nf_layer_submodule.f90
```diff
@@ -269,6 +269,91 @@ impure elemental module subroutine print_info(self)
...
+      type is (reshape3d_layer)
+        ! No parameters to get.
+      class default
+        error stop 'Unknown layer type.'
+    end select
+
+  end subroutine get_params
+
+
+  impure module function set_params(self, params) result(consumed)
```
milancurcic, line 325: @jvo203 Why make `set_params` a function here? A subroutine seems more natural for a method that sets the values in an object.

src/nf/nf_network_submodule.f90
```diff
@@ -369,6 +369,49 @@ module subroutine print_info(self)
...
+
+    ! Check that the number of parameters is correct.
+    if (size(params) .ne. self % get_num_params()) then
+      error stop 'network % set_params: number of parameters does not match.'
+    end if
+
+    consumed = 0
+    do n = 1, size(self % layers)
+      consumed = consumed + self % layers(n) % set_params(params(consumed + 1:))
+    end do
```
milancurcic, line 410: @jvo203 Is `consumed` here simply because `layer % set_params` is currently a function?

src/nf/nf_layer_submodule.f90
jvo203: There are two equivalent ways:

1. set_params is a function that returns the number of parameters "consumed" by this particular layer (in other words moves the array cursor further ahead)

or

2. set_params is a subroutine where the consumed argument is "inout", the number of consumed array entries will be added to the consumed argument, to be used by other layers further down the line

src/nf/nf_network_submodule.f90
jvo203: Yes, set_params is a function right now. But it could be made a subroutine, with the "consumed" as an "inout" parameter updated by each layer, i.e. ```
```
self%layers(n)%set_params(params, consumed)
```

There is more than one way of doing it. I just picked one but surely you can change it in the final version!

milancurcic: Rather than passing all remaining params to each layer and having each layer report back how many it consumed, what do you think about this alternative approach:

In `network % set_params`, which receives the full set of parameters from the user, each iteration determines the start and end index of `params` to pass to `layer % set_params`. Something like this (not tested but should give an idea):

```fortran
    ...
    integer :: nstart, nend
    ...

    nstart = 1
    do n = 1, size(self % layers)
      nend = nstart + self % layers(n) % get_num_params() - 1
      call self % layers(n) % set_params(params(nstart:nend))
      nstart = nend + 1
    end do
```
and in `layer % set_params` (here a subroutine), we'd check that the size of input `params` match the number of params in the layer and `error stop` if not.

I find this a bit cleaner and easier to follow since the `params` slicing is kept to a higher level (in `network` rather than lower levels).

jvo203: OK, I've tested your suggestion in my forked branch. It works fine as it is, the ```nstart``` and ```nend``` are set properly and the layers with ```#params=0``` are handled correctly too. The ```consumed``` variable has been deleted across all the source files, and ```set_parameters(self, params)``` has been converted from a function to a subroutine.
```
      nstart = 1
      do n = 1, size(self%layers)
         nend = nstart + self%layers(n)%get_num_params() - 1
         call self%layers(n)%set_parameters(params(nstart:nend))
         nstart = nend + 1
      end do
```
My only concern is how FORTRAN handles passing arrays with the ending index smaller than the beginning, i.e. ```params(1:0)```, which would happen for the case of zero-parameter layers like inputs etc. Would it not cause undefined behaviour? Anyway, such "degenerate" ```params(1:0)``` is not actually used, it is only passed and no further action is taken within zero-parameter layers, it is simply being ignored. And no segmentation faults happen, the code works as expected.

There is one more change I made inside ```pure module subroutine get_parameters(self, params)``` from ```nf_network_submodule.f90```. There is now an extra line ```if (allocated(params)) deallocate(params)``` before the ```do loop``` starts. This makes the code fool-proof across various unexpected use scenarios:
```
if (allocated(params)) deallocate(params)

      do n = 1, size(self%layers)
         call self%layers(n)%get_parameters(params)
      end do
```

I will **NOT** be submitting a new Pull Request for all these changes as the my forked repository is already out of sync with yours, and there is an added problem with TAB formatting. So if you like your suggested approach with ```params(nstart:nend)``` in ```set_params()``` then please go ahead and apply changes on your side. My side has already been converted to your new approach!

milancurcic: Thank you, this is very helpful! Will incorporate into this PR and add you as co-author.

jvo203: OK. That's great!

milancurcic:
I made `layer % get_num_params` elemental so now `net % layers % get_num_params()` returns an array of integers, each element corresponding to the size of parameters in each of the layers.

jvo203:
Neat, with ```get_num_params``` being elemental there is no need for calling it inside the ```do``` loop for each layer, just one call before ```do``` and the rest is reading each array entry corresponding to each layer. Plus a ```sum()``` to get the total number of parameters.

milancurcic:
I added a few more checks, completed the minimal test suite, and tidied up the example. @jvo203 please let me know if you think any more changes are needed. If not, I'll merge it tomorrow. Thanks again for all your help.

example/get_set_network_params.f90
```diff
@@ -0,0 +1,92 @@
...
+  ypred2 = [(net2 % predict([xtest(i)]), i=1, test_size)]
+
+  print *
+  print '(a)', 'We can also check that the two networks produce the same output:'
+  print *
+  print '("net1 output: ", *(g0,1x))', ypred1
+  print '("net2 output: ", *(g0,1x))', ypred2
+
+  print *
+  print '(a)', 'Original and cloned network outputs match:', all(ypred1 == ypred2)
```
jvo203, line 90: Call it being pedantic or nit-picking but comparing floating-point numbers with ```==``` is fraught with danger. Perhaps a safer choice would be something like ```all(abs(ypred1-ypred2) .lt. epsilon)``` ... Then there is a pesky issue of choosing the "right" ```epsilon``` level.

I wonder how much it is of a problem in this case. Perhaps there is nothing to worry about.

jvo203 (approved):
As long as there is no problem with floating-point comparisons in ```get_set_network_params.f90``` I am happy with it. The ```==``` in ```test_get_set_network_params.f90``` compares integer values (stored in a ```real``` array) so there should not be any rounding errors etc.

example/get_set_network_params.f90
milancurcic: Thanks. It could be a well-put criticism, but I don't think it applies in this case. I'll try to explain. The floating point comparison pitfall usually occurs if we compare the values of a floating point expression with a value that is not exactly representable (e.g. `1.2 + 0.1 == 1.3` is `.false.`). However, there are scenarios where we indeed intend to compare the values exactly, for example when evaluating bit-for-bit reproducible calculations. In this case, we pass exactly the same inputs to two models that have exactly the same parameters, and exactly the same order of operations (because the library code is already compiled with the same compiler and settings and the operations are already ordered internally by the compiler), so I would be surprised that the predictions of the two networks don't evaluate as bit-for-bit exactly equal.

All that said, I didn't intend this to be a rigorous test but an illustration. I did expect the values to be exactly the same, and it's nice that it turns out that they are. I think it's fine to leave it as is, and we can fix it later if a user discovers a setting (hardware+compiler) that makes `all(ypred1 == ypred2)` here `.false.`.

jvo203: Thanks, that's OK then!

PR #123, opened 2023-03-07
Added Leaky ReLU activation function (1D & 3D)

Spnetic-5:
- Solves #46 
@milancurcic

### DIFF
--- a/src/nf/nf_activation_3d.f90
+++ b/src/nf/nf_activation_3d.f90
@@ -20,8 +20,9 @@ module nf_activation_3d
   public :: tanhf, tanh_prime
 
   interface
-    pure function activation_function(x) result(res)
+    pure function activation_function(x, alpha) result(res)
       real, intent(in) :: x(:,:,:)
+      real, intent(in), optional :: alpha
       real :: res(size(x,1),size(x,2),size(x,3))
     end function activation_function
   end interface
@@ -53,51 +54,58 @@ contains
     end where
   end function elu_prime
 
-  pure function exponential(x) result(res)
+  pure function exponential(x, alpha) result(res)
     ! Exponential activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     res = exp(x)
   end function exponential
 
-  pure function gaussian(x) result(res)
+  pure function gaussian(x, alpha) result(res)
     ! Gaussian activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     res = exp(-x**2)
   end function gaussian
 
-  pure function gaussian_prime(x) result(res)
+  pure function gaussian_prime(x, alpha) result(res)
     ! First derivative of the Gaussian activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     res = -2 * x * gaussian(x)
   end function gaussian_prime
 
-  pure function linear(x) result(res)
+  pure function linear(x, alpha) result(res)
     ! Linear activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     res = x
   end function linear
 
-  pure function linear_prime(x) result(res)
+  pure function linear_prime(x, alpha) result(res)
     ! First derivative of the linear activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     res = 1
   end function linear_prime
 
-  pure function relu(x) result(res)
+  pure function relu(x, alpha) result(res)
     !! Rectified Linear Unit (ReLU) activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     res = max(0., x)
   end function relu
 
-  pure function relu_prime(x) result(res)
+  pure function relu_prime(x, alpha) result(res)
     ! First derivative of the Rectified Linear Unit (ReLU) activation function.
     real, intent(in) :: x(:,:,:)
+    real, intent(in), optional :: alpha
     real :: res(size(x,1),size(x,2),size(x,3))
     where (x > 0)
       res = 1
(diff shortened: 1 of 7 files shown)


milancurcic:
Thanks @Spnetic-5! A few things:

* We need a `relu_prime` as well.
* You also need to include it in the select case here:
https://github.com/modern-fortran/neural-fortran/blob/main/src/nf/nf_dense_layer_submodule.f90#L166-L169
and similarly in the nf_conv2d_layer_submodule.f90.
* We shouldn't hardcode alpha (0.01); how best to do it? In the current implementation of how activation functions are passed to the layers, all activation functions need to have the same interface. One approach that I've been thinking about is to make an optional `alpha` parameter in **all** activation functions, except `alpha` would only be used in `leaky_relu` but would be ignored in the other functions. This way the interfaces of all activation functions would match and it should just work.

Spnetic-5:
@milancurcic Please review, I have made suggested changes.

src/nf/nf_dense_layer_submodule.f90
```diff
@@ -5,7 +5,8 @@
                               exponential, &
                               gaussian, gaussian_prime, &
                               linear, linear_prime, &
-                              relu, relu_prime, &
+                              relu, relu_prime, & 
+                              leaky_relu, leaky_relu_prime &
```
milancurcic, line 9:
```suggestion
                              leaky_relu, leaky_relu_prime, &
```

milancurcic:
Thank you, we're almost there. See my 3rd comment from above:

> One approach that I've been thinking about is to make an optional alpha parameter in all activation functions, except alpha would only be used in leaky_relu but would be ignored in the other functions. This way the interfaces of all activation functions would match and it should just work.

The build currently fails because `leaky_relu` has a different interface. Let's make all activation functions take a `real, intent(in), optional :: alpha` argument, to make them have the same interface. This is a somewhat ugly hack, but it should work. You'll also need to add the optional `alpha` to the activation interface here:

https://github.com/modern-fortran/neural-fortran/blob/7039276b0d291d53dc63bc9e9d495192878de864/src/nf/nf_activation_1d.f90#L21-L26

jvdp1:
> The build currently fails because `leaky_relu` has a different interface. Let's make all activation functions take a `real, intent(in), optional :: alpha` argument, to make them have the same interface. This is a somewhat ugly hack, but it should work. You'll also need to add the optional `alpha` to the activation interface here:

Looks good.

Re: `alpha`, wouldn't it be better to define a derived type that contains `alpha`, and pass only the derived type? So, in the future, if other parameters needs to be added for other activation functions, only the content of derived type should be changed and the API would not change.

milancurcic:
Thanks, Jeremie, I think that's a good idea. Having an `activation_params` type would also make it easier to pass this info to individual layers.

Spnetic-5:
> `alpha`, wouldn't it be better to define a derived type that contains `alpha`, and pass only the derived type?

Where and how should I define the derived type, could you please give me example or steps to implement this approach.

Spnetic-5:
@milancurcic @jvdp1 
- By using a derived type to contain the activation parameters, we can add new parameters in the future without changing the function signature. However, this approach does add some overhead for me, since an extra object needs to be created and passed around.
- By adding the optional alpha argument to the activation interface, we allow all the activation functions to have the same interface, which should resolve the issue for now.

As I'm still not familiar with entire Fortran codebase and working, I tried implementing the optional `alpha` approach. Please review the changes made so far; there are significant changes that need to be made in `nf_conv2d_layer_submodule.f90`, but I'm having trouble figuring them out. 
Could you please guide me in the same?

src/nf/nf_activation_3d.f90
```diff
@@ -105,52 +114,79 @@ pure function relu_prime(x) result(res)
...
+    real, intent(in) :: x(:,:,:)
+    real, intent(in) :: alpha
+    real :: res(size(x,1),size(x,2),size(x,3))
+    res = max(alpha*x, x)
+  end function leaky_relu
+
+  pure function leaky_relu_prime(x, alpha) result(res)
+    ! First derivative of the Leaky Rectified Linear Unit (Leaky ReLU) activation function.
+    real, intent(in) :: x(:,:,:)
+    real, intent(in) :: alpha
```
jvdp1, line 128: It should be `optional` everywhere.
In this case, I would use something like the `stdlib` `optval` function.

milancurcic:
I think it's fine that we do just an optional `alpha` as a stopgap solution for this PR, and we can generalize it to `activation_params` in a separate PR.

src/nf/nf_activation_3d.f90
milancurcic: I'd like to evaluate the performance impact of using `optval` since an activation function is invoked many times and are at the very bottom of the call stack. It would be a good opportunity to use stdlib as a dependency, and via fpm it will be easy. But we'd need to add it as a CMake dependency, which will require a bit more CMake code here. I'm wary of making this PR any larger because it's @Spnetic-5's first PR.

Once we do `activation_params` derived type (in a separate PR), there's a technique which will allow us to not rely on `present` or `optval`.

Spnetic-5:
> there are significant changes that need to be made in nf_conv2d_layer_submodule.f90, but I'm having trouble figuring them out.
@milancurcic  Please guide on this

milancurcic:
What needs changing in nf_conv2d_layer_submodule.f90?

Spnetic-5:
> What needs changing in nf_conv2d_layer_submodule.f90?

We need to make changes in these cases 
```
case('elu')
       self % activation => elu
       self % activation_prime => elu_prime
       self % activation_name = 'elu'
```

src/nf/nf_activation_3d.f90
jvdp1: Good points! I agree with you.

milancurcic:
1. Define the `alpha` parameter as an attribute on the `dense_layer` and `conv2d_layer` concrete layer types, e.g. here:

https://github.com/modern-fortran/neural-fortran/blob/7039276b0d291d53dc63bc9e9d495192878de864/src/nf/nf_dense_layer.f90#L15-L33

you can add `real :: alpha`, so that in type-bound methods it will be accessible via `self % alpha`.

2. In the layer constructor functions, rather than passing `alpha` to `set_activation()`, set it to the type attribute `alpha` that you defined in step 1 above (if `present(alpha)`, of course). Add also a run-time check to require `alpha` argument to be present if the `leaky_relu` activation is requested.

3. In forward and backward pass methods for each `dense_layer` and `conv2d_layer`, pass `self % alpha` (which is now set in step 2 in the case of `leaky_relu`) to `self % activation()`.

After you do these steps, I believe it should work correctly.

milancurcic:
To keep this PR small, let's implement the activation params approach in a separate PR. Thanks @Spnetic-5!

PR #126, opened 2023-03-31
Class-based activation functions

ggoyman:
Activation functions are implemented with the use of base abstract class 'activation_function'. 
This new class provides generic `eval` and `eval prime` methods for evaluating activation function and its derivative values (in both 1d and 3d cases). 
Instead of function names, instances of the `activation function` type are passed to the layer constructors:
```fortran
net = network([ &
    input(1), &
    dense(5, activation=sigmoid()), &
    dense(5, activation=elu(alpha=0.1)), &
    dense(1) &
  ])
```


milancurcic:
Thanks @ggoyman! Will review it.

milancurcic:
I added a test to check that the `alpha` is properly set in parametric activation functions. I also set some sane default values for `alpha` from Keras, directly in the type definition.

Thanks again for this great refactor.

milancurcic (approved):

PR #129, opened 2023-05-03
WIP: Update CMake to allow for prebuild dependencies

### LINKED ISSUE #128: Possible to build without internet access aka pre-built dependencies?
A user I help support with the GEOS model has asked for neural-fortran to be used. The main issue I'm currently having is how I can build it in my "usual" way.

For example, currently I use [ESMA-Baselibs](https://github.com/GEOS-ESM/ESMA-Baselibs) to install the "base libraries" used by GEOS. Now, Baselibs is "nice" for clusters in that I can clone it on a node with internet access (and download a few things not on GitHub and not submodule-able) and then do all the other tasks on a compute node where things like `make -j10` are allowed (head nodes maaaaybe they'd allow `make -j2`).

The issue I see here is that if I use the CMake install method, it looks like it *always* uses `FetchContent` which of course would fail on a compute node at CMake time since it couldn't get the code since compute nodes can't see the internet.

So, I wondered, do you have ideas on how to handle this? My first thought was I can add functional-fortran, h5fortran, and json-fortran as submodules, but I think I'd need to do something like in, say, `functional.cmake`:

```cmake
find_package(functional 0.6.1 QUIET)
if (NOT functional_FOUND)

  FetchContent_Declare(functional
    GIT_REPOSITORY https://github.com/wavebitscientific/functional-fortran
    GIT_TAG 0.6.1
    GIT_SHALLOW true
  )

  FetchContent_Populate(functional)
...
```

Does that look about right? I'm going to try this and see in some test builds but I even wondered if you'd be willing to support such a change.

mathomp4:
# WORK IN PROGRESS

This is a work in progress. It works for me but I've currently broken the regular build. Nuts.

---

Closes #128 

This PR is to allow prebuilt functional-fortran, h5fortran, and json-fortran to be used with neural-fortran.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ include(FetchContent)
 include(cmake/options.cmake)
 include(cmake/compilers.cmake)
 
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 include(cmake/functional.cmake)
 include(cmake/h5fortran.cmake)
 include(cmake/json.cmake)
--- a/cmake/Findfunctional.cmake
+++ b/cmake/Findfunctional.cmake
@@ -8,15 +8,12 @@ include(FindPackageHandleStandardArgs)
 
 find_library(functional_LIBRARY
   NAMES functional)
-message(STATUS "functional_LIBRARY: ${functional_LIBRARY}")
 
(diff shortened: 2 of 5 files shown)


mathomp4:
@milancurcic Hmm. I'll take a look.

Though I still think I might need to consult with @scivision on h5fortran. Or, I suppose, maintain a fork of it for now.

Issue #133, opened 2023-05-26
Add example program to fit a quadratic function using different optimizers


milancurcic:
Start with

* Batch gradient descent
* Mini-batch gradient descent
* Stochastic gradient descent

See other simple examples like "simple" and "sine" for reference to get started.

milancurcic:
Done by #134

PR #134, opened 2023-05-26
GSoC Optimizers: Example program to fit a quadratic function

Spnetic-5:
Solving #133 @milancurcic 

**Optimizers to be implemented:**
- [x] Batch gradient descent
- [x] Mini-batch gradient descent
- [x] Stochastic gradient descent

### DIFF
--- a/example/quadratic.f90
+++ b/example/quadratic.f90
@@ -64,7 +64,7 @@ program quadratic_fit
 
 contains
 
-  subroutine sgd_optimizer(net, x, y, learning_rate, num_epochs)
+   subroutine sgd_optimizer(net, x, y, learning_rate, num_epochs)
     ! In the stochastic gradient descent (SGD) optimizer, we run the forward
     ! and backward passes and update the weights for each training sample,
     ! one at a time.
@@ -73,13 +73,19 @@ contains
     real, intent(in) :: learning_rate
     integer, intent(in) :: num_epochs
     integer :: i, n
+    integer, dimension(:), allocatable :: indices
 
     print *, "Running SGD optimizer..."
 
+    ! Generate shuffled indices for the mini-batches
+    allocate(indices(size(x)))
+    indices = [(i, i = 1, size(x))]
+    call shuffle(indices)
+
     do n = 1, num_epochs
       do i = 1, size(x)
-        call net % forward([x(i)])
-        call net % backward([y(i)])
+        call net % forward([x(indices(i))])
+        call net % backward([y(indices(i))])
         call net % update(learning_rate)
       end do
     end do
@@ -130,7 +136,7 @@ contains
     do n = 1, num_epochs
       do j = 1, num_batches
 
-        ! Select the mini batch.
+        ! Select the minibatch.
         start_index = (j - 1) * batch_size + 1
         end_index = j * batch_size
 
@@ -145,5 +151,20 @@ contains
     end do
 
   end subroutine minibatch_gd_optimizer
+
+  subroutine shuffle(arr)
+    ! Shuffles an array using the Fisher-Yates algorithm.
+    integer, dimension(:), intent(inout) :: arr
+    real :: j
+    integer :: i, temp
+
+    do i = size(arr), 2, -1
+      call random_number(j)
+      j = floor(j * real(i)) + 1.0
+      temp = arr(i)
+      arr(i) = arr(int(j))
+      arr(int(j)) = temp
+    end do
+  end subroutine shuffle
 
 end program quadratic_fit
--- a/example/quadratic.f90
+++ b/example/quadratic.f90
@@ -19,7 +19,7 @@ program quadratic_fit
 
   allocate(xtest(test_size), ytest(test_size))
   xtest = [((i - 1) * 2 / test_size, i = 1, test_size)]
-  ytest = (xtest**2 / 2 + xtest / 2 + 1) / 2
+  ytest = quadratic(xtest)
 
   ! x and y as 1-D arrays
   allocate(x(train_size), y(train_size))
@@ -28,8 +28,8 @@ program quadratic_fit
   do i = 1, train_size
     call random_number(x(i))
     x(i) = x(i) * 2
-    y(i) = (x(i)**2 / 2 + x(i) / 2 + 1) / 2
   end do
+  y = quadratic(x)
 
   ! optimizer and learning rate
   learning_rate = 0.01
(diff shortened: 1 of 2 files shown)


milancurcic:
Thanks @Spnetic-5, looks like a good start. You already have the pure SGD example. Do you need any help going forward? To allow batch and mini-batch GDs, I suggest defining `x` and `y` data as 1-d arrays that will be your entire dataset. Then for SGD, feed `x` and `y` elements one at a time, for mini-batch subset multiple batches, and for batch GD pass the entire arrays.

example/quadratic.f90
```diff
@@ -0,0 +1,151 @@
...
+  net = network([ &
+    input(1), &
+    dense(3), &
+    dense(1) &
+  ])
+
+  call net % print_info()
+
+  allocate(xtest(test_size), ytest(test_size), ypred(test_size))
+  xtest = [(i - 1) * 2 / test_size, i = 1, test_size]
```
milancurcic, line 24:
```suggestion
  xtest = [((i - 1) * 2 / test_size, i = 1, test_size)]
```

example/quadratic.f90
```diff
@@ -0,0 +1,151 @@
...
+  ! Mini-batch SGD optimizer
+  call minibatch_sgd_optimizer(net, x, y, learning_rate, num_iterations, batch_size)
+
+  ! Calculate predictions on the test set
+  ypred = [(net % predict([xtest(i)]), i = 1, test_size)]
+
+  ! Print the mean squared error
+  print '(i0,1x,f9.6)', num_iterations, sum((ypred - ytest)**2) / size(ypred)
+
+  deallocate(x, y, xtest, ytest, ypred)
```
milancurcic, line 57: Don't need to deallocate at the end of the program

```suggestion
```

example/quadratic.f90
```diff
@@ -0,0 +1,151 @@
...
+end program quadratic_fit
```
milancurcic, line 58: You may want to put the `end program quadratic_fit` statement at the very end of the program (after subroutine definitions), and place a `contains` here so that the subroutines that follow get the program scope.

example/quadratic.f90
```diff
@@ -0,0 +1,151 @@
...
+
+!   num_layers = net % num_layers()
+
+!   do i = 1, num_layers
+
+!     layer = net % get_layer(i)
+!     layer % weights = layer % weights - learning_rate * layer % gradients
+!     layer % biases = layer % biases - learning_rate * layer % gradients
+!   end do
+! end subroutine update_parameters
```
milancurcic, line 151: I don't think you'll need a custom subroutine for updating here.

I think you can just use the `network % update` method as in

https://github.com/modern-fortran/neural-fortran/blob/8293118220ec1e9c16abda766e3b273f161592b4/src/nf/nf_network_submodule.f90#L553

and in your subroutines that would probably be

```fortran
call net % update(learning_rate / batch_size)
```

milancurcic:
Thanks @Spnetic-5 for the work so far. Please study the changes in https://github.com/modern-fortran/neural-fortran/pull/134/commits/bda1968f70d0cbf03bb275cb0bbb043f74d3b102. There were a few important fixes to the code:

* The training dataset was allocated to the size of the number of epochs (iterations), but the epochs size are really an outer loop to the sample size; I introduced the `train_size` parameter which I use to allocate the training `x` and `y`.
* Not reusing the same network between 3 optimization approaches; each needs its own network instance.
* `net % forward` and `net % backward` methods needed one sample at a time as inputs rather than whole arrays; this is something we can improve later by allowing to pass a batch of data at once.
* Evaluating `ypred` for each optimization method.

I don't know if the results are correct yet, but the code compiles and produces lower errors with increasing epoch count. On my computer the minibatch GD produces very different results between debug and release profiles, so something still not quite correct there.

We're getting close!

Spnetic-5:
You're welcome! I apologize for the errors and the quality of the code pushed earlier. Thank you for pointing out the changes and fixes you made to the code. It's good to hear that the code is now compiling and producing lower errors with increasing epoch count.

I have carefully studied the changes you made, and I understand the modifications you've introduced. It's great to see that the code now compiles without errors and produces lower errors with increasing epoch count. I will continue to review the code and evaluate the results to ensure correctness.

In order to identify the underlying cause and rectify the issue, I will investigate the discrepancies in the minibatch GD results.

These are the results on my PC :

- For 1000 epochs:
```
Stochastic gradient descent MSE: 0.001104
     Batch gradient descent MSE:  0.062504
 Minibatch gradient descent MSE:  0.088675
```

- For 5000 epochs:
```
Stochastic gradient descent MSE: 0.000449
     Batch gradient descent MSE:  0.071504
 Minibatch gradient descent MSE:  0.000996
```

milancurcic:
@Spnetic-5 in SGD subroutine, can you shuffle the mini-batches so that it's truly stochastic? Currently it loops over the mini-batches in the same order every time. Here's my suggested approach:

1. Split the dataset into mini-batches (you already have this);
2. Shuffle the start indices of each mini-batch;
3. Loop over the shuffled start and and indices to extract the desired mini-batch.

The outcome should be that in each epoch the order of mini-batches is random and different. You can take inspiration from

https://github.com/modern-fortran/neural-fortran/blob/8293118220ec1e9c16abda766e3b273f161592b4/src/nf/nf_network_submodule.f90#L532-L535

but even there the mini-batches are not truly shuffled, but rather the start index is randomly selected so that in each epoch there are some data samples that may be unused and there are some that are used more than once.

example/quadratic.f90
```diff
@@ -0,0 +1,170 @@
...
+    real, dimension(:), allocatable :: batch_x, batch_y
+
+    print *, "Running mini-batch GD optimizer..."
+
+    num_samples = size(x)
+    num_batches = num_samples / batch_size
+
+    allocate(batch_x(batch_size), batch_y(batch_size))
+
+    do n = 1, num_epochs
```
milancurcic, line 136: @Spnetic-5 sorry I mistyped the SGD subroutine in my earlier comment. I meant that the mini-batches should be shuffled here in the minibatch GD :).

Spnetic-5: Yeah, even I got confused as there are no batches in SGD, I'll update the changes

milancurcic: It's the same approach as what you already did, but just shuffling the start_indices for mini-batches rather than all indices in the dataset.

Spnetic-5:
@milancurcic I have updated weekly progress on discourse, should I now proceed to next `RMSProp` or `Adam` optimizer, or there are any more changes required in current optimizers

jvdp1:
Thank you @Spnetic-5 for this PR.  Currently this optimizer is only implemented in an example and is not available to other users through the library. Therefore my advice would be to look how to integrate this optimizer in the library. @milancurcic what should be the next step?

milancurcic (approved):
This is in good shape, thanks @Spnetic-5. I'll open an issue later today for next steps.

PR #140, opened 2023-06-14
Example program that shows how to access internal layer parameters

milancurcic:
@Spnetic-5 this PR introduces an example program that shows how to access internal layer parameters (weights and biases). The same approach can be used to access gradients as well, stored as `dw` for weights and `db` for biases.


Spnetic-5:
Thank you @milancurcic for this code, I'll try this in my implementation and will get back to you.

PR #141, opened 2023-06-15
Add missing update for `conv2d_layer`

milancurcic:
Conv2d layers were previously not getting their parameters updated during training. `cnn_mnist` now converges.


milancurcic:
`cnn_mnist` converges with or without this fix but to a relatively low accuracy (~93% in 10 epochs), whereas it should easily get to >98%. While the fix introduced in this PR is necessary for updating `conv2d` layers to work, there is another bug elsewhere that is causing bias and weight gradients to remain zero during training.

PR #143, opened 2023-06-16
add CELU activation function

pablomazo:
Hi.

This PR implements a new activation function, CELU, which i need for a personal project and may be worth having it here.

It follows the [Pytorch implementation](https://pytorch.org/docs/stable/generated/torch.nn.CELU.html) and here is the paper where it is proposed:
https://arxiv.org/abs/1704.07483

Hope this can be helpful.


milancurcic (approved):
Thank you, Pablo! Nice and straightforward addition.

Issue #136, opened 2023-06-13
Implement RMSprop in example/quadratic.f90


milancurcic:
Let's make the first implementation in the example program and then we'll have a better idea how we need to refactor the library to integrate it there.

[Reference](https://en.wikipedia.org/wiki/Stochastic_gradient_descent#RMSProp)

Spnetic-5:
Hello @milancurcic, As you know RMSProp invloves retrieving the `weights` and `gradients` of the network's layers, in order to update the weights based on the gradients. I don't see implementation of these two in `network` code, so how are we planning the implemention exactly.

milancurcic:
Done by #144

Issue #137, opened 2023-06-13
Implement momentum and Nesterov modifications in SGD in example/quadratic.f90


milancurcic:
Similar to #136.

[Reference](https://en.wikipedia.org/wiki/Stochastic_gradient_descent#Momentum)

milancurcic:
Done by #148

PR #144, opened 2023-06-16
Added RMSProp Optimizer subroutine

Spnetic-5:
Solves #136 
This pull request adds an implementation of the RMSprop optimizer subroutine to the existing quadratic example. 

Approach:
- Initialized `rms_weights` and `rms_gradients` arrays of appropriate dimensions.
- Added a nested loop over the network layers to update the weights using the RMSprop update rule.
- Calculated `rms_weights` and `rms_gradients` using the decay rate and current weights/gradients.
- Updated the weights using the RMSprop update rule: `weights = weights - (learning_rate / sqrt(rms_weights + epsilon)) * gradients`.

### DIFF
--- a/example/quadratic.f90
+++ b/example/quadratic.f90
@@ -68,7 +68,7 @@ program quadratic_fit
   ypred_rms_prop = [(net_rms_prop % predict([xtest(i)]), i = 1, test_size)]
 
   ! Print the mean squared error
-  print '(" Stochastic gradient descent MSE:", f9.6)', sum((ypred_sgd - ytest)**2) / size(ytest)
+  print '("Stochastic gradient descent MSE:", f9.6)', sum((ypred_sgd - ytest)**2) / size(ytest)
   print '("     Batch gradient descent MSE: ", f9.6)', sum((ypred_batch_sgd - ytest)**2) / size(ytest)
   print '(" Minibatch gradient descent MSE: ", f9.6)', sum((ypred_minibatch_sgd - ytest)**2) / size(ytest)
   print '("                    RMSProp MSE: ", f9.6)', sum((ypred_rms_prop - ytest)**2) / size(ytest)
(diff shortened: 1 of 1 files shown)


Spnetic-5:
Apologies for late reply @milancurcic

milancurcic:
@Spnetic-5 I mostly re-wrote the subroutine so that now it compiles and converges. It's not using mini-batching; for simplicity, for now it's being applied after the entire batch of forward and backward passes.

I understand that this PR was challenging. It took me a bit to find the right approach. In your most recent commit, you made some changes and wrote "made suggested corrections", which made it sound like the PR was good to go. However, the example was not even compiling at this stage. Whenever you struggle with the implementation, please write a comment in the PR explaining where you got stuck and if you need help, rather than just leaving it with a short commit message.

Also please study the implementation in this PR. It introduces a new derived type to allow tracking a moving average of gradients over multiple epochs and for each layer. We are likely to use this approach for other optimizers that need a moving average logic.

Spnetic-5:
I apologize for the confusion caused by my commit message. It was not my intention to imply that the code was ready to go. But the code was compiling and running well in my PC, I'll make sure to provide detailed comments in the pull request in the future.

milancurcic:
Thank you, @Spnetic-5, and no worries. I apologize for jumping the gun and finishing the implementation in this PR.

Going forward, would you like to give it a shot to continue the work in #139, or would you like to implement another optimizer in the quadratic fit example program? Recall that once we implement #139 for SGD, the new optimizers in quadratic will serve as prototype implementations for porting them into the library.

Spnetic-5:
> Going forward, would you like to give it a shot to continue the work in #139, or would you like to implement another optimizer in the quadratic fit example program? Recall that once we implement #139 for SGD, the new optimizers in quadratic will serve as prototype implementations for porting them into the library.

Thank you, @milancurcic. I would like to work on #137 first. Once we have completed that, we can move on to #139 and then additional new optimizers.

PR #150, opened 2023-07-18
Added Adam optimizer implementation

Spnetic-5:
#### Adam Implementation:

- Momentum Update:
`m = beta1 * m + (1 - beta1) * gradient`

- Adaptive Learning Rates Update:
`v = beta2 * v + (1 - beta2) * gradient**2`

- Bias-Corrected First and Second Moment Estimates:
`m_hat = m / (1 - beta1**t)`
`v_hat = v / (1 - beta2**t)`

- Parameter Update:
`param = param - learning_rate * m_hat / (sqrt(v_hat) + epsilon)`

### DIFF
--- a/src/nf/nf_optimizers.f90
+++ b/src/nf/nf_optimizers.f90
@@ -166,23 +166,14 @@ contains
     class(adam), intent(inout) :: self
     real, intent(inout) :: param(:)
     real, intent(in) :: gradient(:)
-    real, allocatable :: reg_gradient(:)
 
     self % t = self % t + 1
 
-    if (self % weight_decay > 0.0) then
-      ! Compute the modified gradient with weight decay regularization
-      allocate(reg_gradient(size(param)))
-      reg_gradient = gradient + 2.0 * self % weight_decay * param
-      ! Update biased first moment estimate.
-      self % m = self % beta1 * self % m + (1 - self % beta1) * reg_gradient
-      ! Update biased second raw moment estimate.
-      self % v = self % beta2 * self % v + (1 - self % beta2) * reg_gradient**2
-    else
-      self % m = self % beta1 * self % m + (1 - self % beta1) * gradient
-      ! Update biased second raw moment estimate.
-      self % v = self % beta2 * self % v + (1 - self % beta2) * gradient**2
-    end if
+    ! Update biased first moment estimate.
+    self % m = self % beta1 * self % m + (1 - self % beta1) * gradient
+
+    ! Update biased second raw moment estimate.
+    self % v = self % beta2 * self % v + (1 - self % beta2) * gradient**2
 
     ! Compute bias-corrected first and second moment estimates.
     associate( &
@@ -190,9 +181,9 @@ contains
       v_hat => self % v / (1 - self % beta2**self % t) &
     )
 
-      ! Update parameters.
-      param = param - self % learning_rate &
-        * m_hat / (sqrt(v_hat) + self % epsilon)
+    ! Update parameters.
+    param = param - self % learning_rate &
+      * m_hat / (sqrt(v_hat) + self % epsilon) - self % weight_decay * param
 
     end associate
 
--- a/src/nf/nf_optimizers.f90
+++ b/src/nf/nf_optimizers.f90
@@ -58,7 +58,7 @@ module nf_optimizers
     !! http://www.cs.toronto.edu/~hinton/coursera/lecture6/lec6.pdf
     real :: decay_rate = 0.9
     real :: epsilon = 1e-8
-    real, allocatable :: rms_gradient(:)
+    real, allocatable, private :: rms_gradient(:)
   contains
     procedure :: init => init_rmsprop
     procedure :: minimize => minimize_rmsprop
@@ -73,7 +73,7 @@ module nf_optimizers
     real :: beta1 = 0.9
     real :: beta2 = 0.999
     real :: epsilon = 1e-8
-    real, allocatable :: m(:), v(:)
+    real, allocatable, private :: m(:), v(:)
     integer :: t = 0
   contains
     procedure :: init => init_adam
--- a/src/nf/nf_optimizers.f90
+++ b/src/nf/nf_optimizers.f90
@@ -43,7 +43,7 @@ module nf_optimizers
     !! Stochastic Gradient Descent optimizer
     real :: momentum = 0
     logical :: nesterov = .false.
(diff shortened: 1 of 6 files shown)


src/nf/nf_optimizers.f90
```diff
@@ -123,4 +135,47 @@ pure subroutine minimize_rmsprop(self, param, gradient)
...
+    class(adam), intent(inout) :: self
+    integer, intent(in) :: num_params
+    if (.not. allocated(self % m)) then
+      allocate(self % m(num_params))
+      self % m = 0
+    end if
+    if (.not. allocated(self % v)) then
+      allocate(self % v(num_params))
+      self % v = 0
+    end if
```
milancurcic, line 149: `m` and `v` will always be either both allocated or deallocated so we can fold the second `if` block into the first one.

```suggestion
    if (.not. allocated(self % m)) then
      allocate(self % m(num_params), self % v(num_params))
      self % m = 0
      self % v = 0
    end if
```

src/nf/nf_optimizers.f90
```diff
@@ -59,6 +59,18 @@ end subroutine minimize
     procedure :: minimize => minimize_rmsprop
   end type rmsprop
 
+  type, extends(optimizer_base_type) :: adam
+    !! Adam optimizer
```
milancurcic, line 63: Let's add the reference to the paper here.

milancurcic (commented):
Thank you! Good so far.

milancurcic (approved):
Thank you @Spnetic-5, this is almost good to go. The only thing missing is the convergence test for Adam in test/test_optimizers.f90.

src/nf/nf_optimizers.f90
```diff
@@ -123,4 +144,44 @@ pure subroutine minimize_rmsprop(self, param, gradient)
 
   end subroutine minimize_rmsprop
 
+
+  impure elemental subroutine init_adam(self, num_params)
+    class(adam), intent(inout) :: self
+    integer, intent(in) :: num_params
+    if (.not. allocated(self % m)) then
```
jvdp1, line 151: Could `v` be allocated and not `m`?
I suggest to check the status of both `m` and `v`.

milancurcic: That would be an error state, see https://github.com/modern-fortran/neural-fortran/pull/150#discussion_r1266806684. `v` and `m` will always have the same allocation status.

@Spnetic-5 let's make `m` and `v` `private` so that the user can't fiddle with them directly.

I'm also thinking whether we should make the `init` method `private` and invoke it from `minimize`. This decreases the number of places in the code at the network level that the optimizer instance is modified.

milancurcic:
@Spnetic-5, just checking, your latest commit message mentions the convergence test but the commit doesn't actually include the test.

Spnetic-5:
> @Spnetic-5, just checking, your latest commit message mentions the convergence test but the commit doesn't actually include the test.

Sorry, forgot to push test file😅️

src/nf/nf_optimizers.f90
```diff
@@ -123,4 +150,45 @@ pure subroutine minimize_rmsprop(self, param, gradient)
...
+    !! Concrete implementation of an Adam optimizer update rule.
+    class(adam), intent(inout) :: self
+    real, intent(inout) :: param(:)
+    real, intent(in) :: gradient(:)
+
+    self % t = self % t + 1
+
+    ! If weight_decay > 0, use L2 regularization;
+    ! otherwise, default to regular Adam.
+    associate(g => gradient + 2 * self % weight_decay * param)
```
milancurcic, line 175: Why do you have a factor of 2 here? In the [paper, Algorithm 2, line 6](https://arxiv.org/pdf/1711.05101.pdf) I see just $\lambda \theta_{t-1}$ (i.e. `self % weight_decay * param`).

Another question is, based on my understanding from the paper, applying `weight_decay * param` here is L2 regularization. However, for decoupled weight decay regularization, it seems like it should be added below on L188.

src/nf/nf_optimizers.f90
```diff
@@ -123,4 +150,45 @@ pure subroutine minimize_rmsprop(self, param, gradient)
...
+
+    ! Compute bias-corrected first and second moment estimates.
+    associate( &
+      m_hat => self % m / (1 - self % beta1**self % t), &
+      v_hat => self % v / (1 - self % beta2**self % t) &
+    )
+
+      ! Update parameters.
+      param = param - self % learning_rate &
+        * m_hat / (sqrt(v_hat) + self % epsilon)
```
milancurcic, line 188: Following up on my comment above, should the decoupled weight regularization not be:

```suggestion
      param = param - self % learning_rate &
        * m_hat / (sqrt(v_hat) + self % epsilon) &
        - self % weight_decay * param
```

src/nf/nf_optimizers.f90
Spnetic-5: You're right, the scaling factor of 2 is not present in the paper. The reason was when the weight decay coefficient is large, the term `weight_decay * param` can become very large.

Please review the recent changes.

src/nf/nf_optimizers.f90
Spnetic-5: Yeah, have made that change, just pushed

src/nf/nf_optimizers.f90
milancurcic: Thanks @Spnetic-5. We still need to decide how to handle the two different regularization approaches (L2 and decoupled). We currently have both enabled if `weight_decay > 0`. However, I think only one or the other should be applied at a time. PyTorch and Keras differentiate the two by providing two separate classes: Adam and AdamW. I think this is unnecessary in our case. We can have just Adam for now, and allow two different regularization approaches to be applied via input arguments. What do you think about using two separate weight decays to control this, i.e.:

```fortran
  real :: weight_decay_l2 = 0  ! L2 regularization (Adam)
  real :: weight_decay_decoupled = 0 ! decoupled weight decay regularization (AdamW)
```

Spnetic-5: Yup, I agree that having two separate input `weight_decay` arguments isa good approach. So by default, both weight decays will be set to 0, so no regularization will be applied. I'll go ahead and do the changes.

milancurcic:
Thank you, @Spnetic-5, will merge. Please release v0.14.0 when ready.

PR #154, opened 2023-07-28
Adagrad Optimizer Implementation

Spnetic-5:
Reference: [PyTorch Docs]( https://pytorch.org/docs/stable/generated/torch.optim.Adagrad.html#torch.optim.Adagrad)

### DIFF
--- a/src/nf/nf_optimizers.f90
+++ b/src/nf/nf_optimizers.f90
@@ -95,9 +95,9 @@ module nf_optimizers
     !! of Machine Learning Research, 12(Jul), pp.2121-2159.
     !! http://www.jmlr.org/papers/volume12/duchi11a/duchi11a.pdf
     real :: epsilon = 1e-8
-    real, allocatable, private :: sum_squared_gradient(:)
     real :: weight_decay_l2 = 0
     real :: learning_rate_decay = 0
+    real, allocatable, private :: sum_squared_gradient(:)
     integer, private :: t = 0
   contains
     procedure :: init => init_adagrad
@@ -210,6 +210,7 @@ contains
 
   end subroutine minimize_adam
 
+
   impure elemental subroutine init_adagrad(self, num_params)
     class(adagrad), intent(inout) :: self
     integer, intent(in) :: num_params
@@ -218,6 +219,7 @@ contains
       self % sum_squared_gradient = 0
     end if
   end subroutine init_adagrad
+
 
   pure subroutine minimize_adagrad(self, param, gradient)
     !! Concrete implementation of an Adagrad optimizer update rule.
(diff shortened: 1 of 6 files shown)


src/nf/nf_optimizers.f90
```diff
@@ -87,6 +87,20 @@ end subroutine minimize
...
 
+  type, extends(optimizer_base_type) :: adagrad
+    !! Adagrad optimizer by Duchi et al. (2011)
+    !!
+    !! Duchi, J., Hazan, E. and Singer, Y., 2011. Adaptive subgradient
+    !! methods for online learning and stochastic optimization. Journal
+    !! of Machine Learning Research, 12(Jul), pp.2121-2159.
+    !! http://www.jmlr.org/papers/volume12/duchi11a/duchi11a.pdf
+    real :: epsilon = 1e-8
+    real, allocatable, private :: sum_squared_gradient(:)
```
milancurcic, line 98: Need `learning_rate_decay` here ($\eta$ in https://pytorch.org/docs/stable/generated/torch.optim.Adagrad.html#torch.optim.Adagrad)

src/nf/nf_optimizers.f90
```diff
@@ -187,10 +201,34 @@ pure subroutine minimize_adam(self, param, gradient)
     ! Update parameters.
     param = param &
       - self % learning_rate * m_hat / (sqrt(v_hat) + self % epsilon) &
-      - self % weight_decay_decoupled * param
+      - self % learning_rate * self % weight_decay_decoupled * param
```
milancurcic, line 204: I suggest:

```
param = param &
  - self % learning_rate * (m_hat / (sqrt(v_hat) + self % epsilon) + self % weight_decay_decoupled * param)
```

src/nf/nf_optimizers.f90
```diff
@@ -187,10 +201,34 @@ pure subroutine minimize_adam(self, param, gradient)
...
+    !! Concrete implementation of an Adagrad optimizer update rule.
+    class(adagrad), intent(inout) :: self
+    real, intent(inout) :: param(:)
+    real, intent(in) :: gradient(:)
+
+    ! Update the sum of squared gradients using the Adagrad rule
+    self % sum_squared_gradient = self % sum_squared_gradient + gradient**2
+
+    ! Update the network parameters based on the new sum of squared gradients
+    param = param - self % learning_rate &
```
milancurcic, line 229: The learning rate needs to adapt with every time step here, according to https://pytorch.org/docs/stable/generated/torch.optim.Adagrad.html#torch.optim.Adagrad.

src/nf/nf_optimizers.f90
milancurcic: Both Keras and PyTorch implement the `weight_decay` (the L2 kind) in Adagrad. We may as well do that since we're already here.

milancurcic:
Thanks @Spnetic-5. I believe it's correct now. In your original implementation, the L2 regularization was not accounted for in the accumulation of the squared gradients because you applied it later in the param update. The learning rate decay was also doubly accounted for because in each step the learning rate should be amortized relative to the original learning rate, not the one from the previous step. Subtle differences that weren't caught in the tests.

I'll go ahead and merge, please release v0.15.0 when you get a chance.

Issue #155, opened 2023-08-08
Implement `batchnorm` layer


milancurcic:
Originally requested by @rweed in #114.

A batch normalization is possibly the next most widely used layer after dense, convolutional, and maxpooling layers, and is an important tool in optimization (accelerating training).

For neural-fortran, it will mean that we will need to allow passing a batch of data to individual layer's `forward` and `backward` methods. While for `dense` and `conv2d` layers it may also mean an opportunity to numerically optimize the operations (e.g. running the same operation on a batch of data instead of over one sample at a time), for a batchnorm layer it is required because this layer evaluates moments (e.g. means and standard deviations) over a batch of inputs to normalize the input data.

Implementing batchnorm will require another non-trivial refactor like we did to enable generic optimizers. However, it will probably be easier. The first step will be to allow the passing of a batch of data to `forward` and `backward` methods, as I mentioned above. In other words, this snippet:

https://github.com/modern-fortran/neural-fortran/blob/b119194a6472e1759966ef9b0ee02dc66ddb1a3a/src/nf/nf_network_submodule.f90#L587-L590

after the refactor, we should be able to write like this:

```fortran
call self % forward(input_data(:,:))
call self % backward(output_data(:,:))
```

where the first dim corresponds to inputs and outputs in input and output layers, respectively, and the second dim corresponds to multiple samples in a batch. I will open a separate issue for this.

@Spnetic-5 given limited time in the remainder of the GSoC program, we may be unable to complete the batchnorm implementation, but we can make significant headway on it for sure.

## References

* [Ioffe and Szegedy (2015)](https://arxiv.org/abs/1502.03167)
* [`BatchNormalization` in Keras](https://keras.io/api/layers/normalization_layers/batch_normalization/)
* [`BatchNorm1d` in PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html#torch.nn.BatchNorm1d)

Issue #156, opened 2023-08-08
Refactor `forward` and `backward` methods to allow passing a batch of data instead of one sample at a time


milancurcic:
In support of #155.

This will impact the `forward` and `backward` methods in:

* `network` type
* `layer` type
* `dense_layer` type
* `conv2d_layer` type

Effectively, rather than looping over sample in a batch inside of `network % train`, we will pass batches of data all the way down to the lowest level, that is, the `forward` and `backward` methods of `dense_layer` and `conv2d_layer` types. Lowering the looping over the sample in a batch will also allow the implementation of a `batchnorm_layer`.

It will also potentially allow more efficient `matmul`s in dense and conv layers if we replace the stock `matmul` with some more specialized and efficient `sgemm` or similar from some flavor of BLAS or MKL.

PR #157, opened 2023-08-10
Added Batch Normalization Layer modules

Spnetic-5:
Addresses #155  

@milancurcic , I've included the structure of the batch normalization layer. Could you please review it and confirm whether I'm in the correct direction?

### DIFF
--- a/src/nf/nf_batch_norm_layer.f90
+++ b/src/nf/nf_batch_norm_layer.f90
@@ -1,0 +1,108 @@
+module nf_batch_norm_layer
+
+  !! This module provides a batch normalization `batch_norm_layer` type.
+
+  use nf_base_layer, only: base_layer
+  implicit none
+
+  private
+  public :: batch_norm_layer
+
+  type, extends(base_layer) :: batch_norm_layer
+
+    integer :: size
+    real, allocatable :: gamma(:)
+    real, allocatable :: beta(:)
+    real, allocatable :: running_mean(:)
+    real, allocatable :: running_var(:)
+    real, allocatable :: input(:,:)
+    real, allocatable :: output(:,:)
+    real, allocatable :: gamma_grad(:)
+    real, allocatable :: beta_grad(:)
+    real, allocatable :: input_grad(:,:)
+
+  contains
+
+    procedure :: forward
+    procedure :: backward
+    procedure :: get_gradients
+    procedure :: get_num_params
+    procedure :: get_params
+    procedure :: init
+    procedure :: set_params
+
+  end type batch_norm_layer
+
+  interface batch_norm_layer
+    pure module function batch_norm_layer_cons(size) result(res)
+      !! `batch_norm_layer` constructor function
+      integer, intent(in) :: size
+      type(batch_norm_layer) :: res
+    end function batch_norm_layer_cons
+  end interface batch_norm_layer
+
+  interface
+
+    module subroutine init(self, input_shape)
+      !! Initialize the layer data structures.
+      !!
+      !! This is a deferred procedure from the `base_layer` abstract type.
+      class(batch_norm_layer), intent(in out) :: self
+        !! A `batch_norm_layer` instance
+      integer, intent(in) :: input_shape(:)
+        !! Input layer dimensions
+    end subroutine init
+
+    pure module subroutine forward(self, input)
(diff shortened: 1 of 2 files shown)


milancurcic:
Thank you @Spnetic-5, great. Yes, this is certainly in the correct direction. I haven't yet reviewed the forward and backward algorithms, let me know when it's ready for review.

Also, despite #156, we should be able to test this layer implementation on its own, without integrating it with the network. In other words, while #156 will be necessary for full integration, we can work on this implementation as a standalone layer before #156.

Spnetic-5:
@milancurcic  Please review the forward and backward pass implementations I've added based on my interpretation of the paper, also could you guide me on how we can test this layer?

milancurcic:
Thanks! Let's review it briefly on the call today.

We can test this layer independently by passing some small, known input, and comparing the result with the corresponding known output. This should be straightforward since the batchnorm operation is relatively simple (just normalization of data). The backward is the inverse operation, so as I understand it, we can pass the same expected output to recover the same expected input.

milancurcic:
@Spnetic-5 I just saw your message on Discourse, no problem; we'll proceed work on this PR as usual.

milancurcic:
See for example a program that tests forward and backward passes of the maxpool2d layer using known inputs and expected outputs:

https://github.com/modern-fortran/neural-fortran/blob/main/test/test_maxpool2d_layer.f90

We'd use the same approach to test a batchnorm layer.

Spnetic-5:
Hello, @milancurcic. Sorry for the lack of activity over the past few days; this was my final week of internship in Canada, and I'll be returning to India on Monday.

I added a test module for the batch norm layer, however it has some error; I believe I will need your assistance on this.

milancurcic:
No worries at all, thanks for all the work. I'll review it tomorrow.

src/nf/nf_batchnorm_layer_submodule.f90
```diff
@@ -0,0 +1,105 @@
...
+    integer, intent(in) :: num_features
+    type(batchnorm_layer) :: res
+
+    res % num_features = num_features
+    allocate(res % gamma(num_features), source=1.0)
+    allocate(res % beta(num_features))
+    allocate(res % running_mean(num_features), source=0.0)
+    allocate(res % running_var(num_features), source=1.0)
+    allocate(res % input(num_features, num_features))
+    allocate(res % output(num_features, num_features))
```
milancurcic, line 18: I don't think the shape for these should be `(num_features, num_features)`, but rather `(batch_size, num_features)`. The `batch_size` also won't be known until the first forward pass, so we should defer the allocation until then. In the forward pass, we could have a simple `allocated` check to see if they have not been allocated then, and allocate them to the shape of the `input`.

src/nf/nf_batchnorm_layer_submodule.f90
```diff
@@ -0,0 +1,105 @@
...
+    class(batchnorm_layer), intent(in out) :: self
+    real, intent(in) :: input(:,:)
+
+    ! Store input for backward pass
+    self % input = input
+
+    associate( &
+      ! Normalize the input
+      normalized_input => (input - reshape(self % running_mean, shape(input, 1))) &
+        / sqrt(reshape(self % running_var, shape(input, 1)) + self % epsilon) &
```
milancurcic, line 52: `running_mean` and `running_var` are not yet updated anywhere, only initialized.

test/test_batchnorm_layer.f90
```diff
@@ -0,0 +1,65 @@
...
+    write(stderr, '(a)') 'batchnorm layer should not be marked as initialized yet.. failed'
+  end if
+
+  ! Initialize sample input and gradient
+  allocate(sample_input(num_features, 1))
+  allocate(gradient(num_features, 1))
+  sample_input = 1.0
+  gradient = 2.0
+
+  !TODO run forward and backward passes directly on the batchnorm_layer instance
```
milancurcic, line 37: I'll add a simple test directly on the `batchnorm_layer` instance rather than the high-level `layer_type` instance so you get the idea how it will work.

src/nf/nf_batchnorm_layer_submodule.f90
milancurcic: Sorry, I meant `(num_features, batch_size)`.

PR #162, opened 2023-10-28
Implementing basic RNN

castelao:
A work in progress. I'm mostly interested in loading a TF model from HDF and applying predict(), but I'll do my best in doing a complete implementation coherent with the rest of the library.

### DIFF
--- a/src/nf/nf_layer_submodule.f90
+++ b/src/nf/nf_layer_submodule.f90
@@ -296,7 +296,7 @@ contains
       type is (rnn_layer)
         num_params = this_layer % get_num_params()
       class default
-        error stop 'Unknown layer type.'
+        error stop 'get_num_params() with unknown layer type.'
     end select
 
   end function get_num_params
@@ -323,7 +323,7 @@ contains
       type is (rnn_layer)
         params = this_layer % get_params()
       class default
-        error stop 'Unknown layer type.'
+        error stop 'get_params() with unknown layer type.'
     end select
 
   end function get_params
@@ -348,7 +348,7 @@ contains
       type is (reshape3d_layer)
         ! No gradients to get.
       class default
-        error stop 'Unknown layer type.'
+        error stop 'get_gradients() with unknown layer type.'
     end select
 
   end function get_gradients
@@ -405,7 +405,7 @@ contains
           // 'on a zero-parameter layer; nothing to do.'
 
           class default
-        error stop 'Unknown layer type.'
+        error stop 'set_params() with unknown layer type.'
     end select
 
   end subroutine set_params
--- a/src/nf/nf_layer_submodule.f90
+++ b/src/nf/nf_layer_submodule.f90
@@ -8,6 +8,7 @@ submodule(nf_layer) nf_layer_submodule
   use nf_input3d_layer, only: input3d_layer
   use nf_maxpool2d_layer, only: maxpool2d_layer
   use nf_reshape_layer, only: reshape3d_layer
+  use nf_rnn_layer, only: rnn_layer
   use nf_optimizers, only: optimizer_base_type
 
 contains
@@ -292,6 +293,8 @@ contains
         num_params = 0
       type is (reshape3d_layer)
         num_params = 0
+      type is (rnn_layer)
+        num_params = this_layer % get_num_params()
       class default
         error stop 'Unknown layer type.'
     end select
--- a/src/nf/nf_layer_submodule.f90
+++ b/src/nf/nf_layer_submodule.f90
@@ -320,6 +320,8 @@ contains
         ! No parameters to get.
       type is (reshape3d_layer)
         ! No parameters to get.
+      type is (rnn_layer)
+        params = this_layer % get_params()
       class default
         error stop 'Unknown layer type.'
     end select
--- a/src/nf/nf_rnn_layer.f90
+++ b/src/nf/nf_rnn_layer.f90
@@ -37,7 +37,7 @@ module nf_rnn_layer
     !procedure :: forward
     !procedure :: get_gradients
     procedure :: get_num_params
-    !procedure :: get_params
+    procedure :: get_params
     procedure :: init
     !procedure :: set_params
 
--- a/src/nf/nf_rnn_layer.f90
+++ b/src/nf/nf_rnn_layer.f90
@@ -34,7 +34,7 @@ module nf_rnn_layer
   contains
 
     !procedure :: backward
-    !procedure :: forward
+    procedure :: forward
     !procedure :: get_gradients
     procedure :: get_num_params
     procedure :: get_params
--- a/src/nf/nf_rnn_layer.f90
+++ b/src/nf/nf_rnn_layer.f90
@@ -1,0 +1,131 @@
+module nf_rnn_layer
+
+  !! This module provides the concrete dense layer type.
+  !! It is used internally by the layer type.
+  !! It is not intended to be used directly by the user.
+
+  use nf_activation, only: activation_function
+  use nf_base_layer, only: base_layer
+
+  implicit none
+
+  private
+  public :: rnn_layer
+
+  type, extends(base_layer) :: rnn_layer
+
+    !! Concrete implementation of a dense (fully-connected) layer type
+
+    integer :: input_size
+    integer :: output_size
+
+    real, allocatable :: weights(:,:)
+    real, allocatable :: recurrent(:,:)
+    real, allocatable :: biases(:)
+    real, allocatable :: state(:)
+    real, allocatable :: z(:) ! matmul(x, w) + b
+    real, allocatable :: output(:) ! activation(z)
+    real, allocatable :: gradient(:) ! matmul(w, db)
+    real, allocatable :: dw(:,:) ! weight gradients
+    real, allocatable :: db(:) ! bias gradients
+
+    class(activation_function), allocatable :: activation
+
+  contains
+
+    !procedure :: backward
+    !procedure :: forward
+    !procedure :: get_gradients
+    procedure :: get_num_params
+    !procedure :: get_params
+    procedure :: init
+    !procedure :: set_params
+
+  end type rnn_layer
+
+  interface rnn_layer
+    elemental module function rnn_layer_cons(output_size, activation) &
+      result(res)
+      !! This function returns the `dense_layer` instance.
+      integer, intent(in) :: output_size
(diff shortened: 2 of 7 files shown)


milancurcic:
Amazing, thanks and great to see you, @castelao, after so many years. 🙂

castelao:
@milancurcic , yes, it's great to connect again. Thank you and the other developers for your time in this library! It is great.

I'm not fluent in modern Fortran, so if you see anything that doesn't make sense, please let me know. And be aware, it is still a WIP.

src/nf/nf_rnn_layer_submodule.f90
```diff
@@ -0,0 +1,169 @@
...
+    self % biases = 0
+
+    allocate(self % output(self % output_size))
+    self % output = 0
+
+    allocate(self % z(self % output_size))
+    self % z = 0
+
+    allocate(self % state(self % output_size))
+    self % state = 0
```
castelao, line 156: @milancurcic , I have a question for you. I'm implementing just a basic RNN with no warm-up. For every new sequence, this state should be reset to some initial value, let's say zero. There are two options here:
- The inputs could be a 2D array, sequence x feature. I tried to implement it, but I think I was breaking logical coherence with the 3D family of procedures.
- Some 'reset()' procedure to avoid needing to re-instantiate the net for every new sequence.

What do you suggest?

A note, I just realized that predict() is based on forward(), thus it necessarily update the state while predicting.

milancurcic: State is what gives the RNN layer memory, correct? So, we want to reset the state between each sequence in both training and inference, right? Calling `init` doesn't work because we only want to reset `state`, but not the learned parameters of the layer. If I got this correctly and you agree, then I think a `reset()` subroutine that is specific to the RNN layer makes perfect sense.

For the first question

> The inputs could be a 2D array, sequence x feature. I tried to implement it, but I think I was breaking logical coherence with the 3D family of procedures.

It can be done (and we should, eventually), but for simplicity and to keep this PR manageable, let's for now stick with the simple case of 1-d input (not sure if that's one value of each feature or a sequence of one feature; you let me know, I've never used or studied much the recurrent layers).

> A note, I just realized that predict() is based on forward(), thus it necessarily update the state while predicting.

Yes, `predict` is just a convenience function wrapper around the `forward` subroutine to return the output of the network.

castelao: Yes and Yes.

Sure, `reset()` sounds like a good way. Good point on keeping the PR small. I think it would be more powerful to allow multiple features, so this 1D would be features, and the sequence evolution can be obtained with multiple calls on `forward()` or `predict()` since the state is preserved.

Thanks!

castelao: @milancurcic , could you take a look, please, in the new `reset()`? It wasn't clear to me what defines a pure/impure module.

I didn't realized at first, but it should be accessible in the top net level since if can be composed with multiple RNN layers.

I haven't included in the PR yet, but it pass my local tests.

milancurcic: On second thought, do you think the resetting of the RNN layer state can be done automatically by the library code, or does it have to be done by the user? It seems to me a but unwieldy to have `nework % reset()` that's specific to recurrent layers only. I see that Keras allows `initial_state` to be passed to the layer (and initializes to zero by default): https://keras.io/api/layers/recurrent_layers/simple_rnn/.

Do you think it would be a good approach to define the following subroutines:

in `nf_layer.f90`

```fortran
subroutine set_state(self, state)
  type(layer), intent(inout) :: self
  real, intent(in), optional :: state(:)
  if (present(state)) then
    ! set the state on the rnn_layer concrete instance
  else
    ! set the state to zero on the rnn_layer concrete instance
  end if
  ...
end subroutine set_state
```

and in `nf_rnn_layer.f90`:

```fortran
subroutine set_state(self, state)
  type(rnn_layer), intent(inout) :: self
  real, intent(in), optional :: state(:)
  if (present(state)) then
    self % state = state
  else
    self % state = 0
  end if
end subroutine set_state
```

But not defining it on the network level, but rather setting it directly on the layers (this would be in user's application code):

```fortran
! Reset the state on all recurrent layers
do n = 1, size(net % layers)
  if (net % layers(n) % name == 'rnn') call net % layers(n) % set_state()
end do
```

castelao: I see your point. Adding a reset in the top level that would only be useful for RNNs is something.

Yes, that should work. Eventually the N-D array input where we can define, or have a convention on, which dimension is the sequence would be the best since it could reset itself.

I'll update the PR with your suggestion. Thanks for your time.

castelao: @milancurcic , is this close to what you had in mind?

milancurcic: Yes, excellent! Do you think this is ready for me to start playing with locally?

castelao: It is missing a few things, including better tests. I also think that I messed up in the back propagation of the error. I'll check that on the first chance. I've been calibrating my NN outside, with TensorFlow itself.

But, if you have the time, it would be great if you could check it sooner than later for logical errors. I don't know modern fortran.

milancurcic:
I added the support for `rnn_layer` to `network % predict` so that we can run the simple_rnn example. However I can't get the simple example to converge and I'm not very familiar with recurrent networks. @castelao   is there a small toy example like the one in simple.f90 that's known to converge and fast? We can then use it for testing as well. With a working example and a small test suite (I can add this), this PR is almost good to go.

castelao:
@milancurcic , I have to check how you updated the library since the last time I worked on this and see if what I did is still consistent. If it looks fine, I intend to work on the following:
- The toy example
- Load functionality from an HDF exported by TensorFlow. I want to avoid recompiling every time I change my coefficients.

Are there any other requirements before I can submit this PR for review? It will be slow progress, but I'm back to this.

src/nf/nf_layer_submodule.f90
```diff
@@ -399,10 +438,27 @@ module subroutine set_params(self, params)
...
+        call this_layer % set_params(params)
+
           class default
-        error stop 'Unknown layer type.'
+        error stop 'set_params() with unknown layer type.'
     end select
 
   end subroutine set_params
 
+  module subroutine set_state(self, state)
```
jvdp1, line 450:
```suggestion
  pure module subroutine set_state(self, state)
```

src/nf/nf_rnn_layer.f90
```diff
@@ -0,0 +1,137 @@
...
+  contains
+
+    procedure :: backward
+    procedure :: forward
+    procedure :: get_gradients
+    procedure :: get_num_params
+    procedure :: get_params
+    procedure :: init
+    procedure :: set_params
+    ! procedure :: set_state
```
jvdp1, line 43:
```suggestion
```

src/nf/nf_rnn_layer.f90
```diff
@@ -0,0 +1,137 @@
...
+        !! Dense layer instance
+      integer, intent(in) :: input_shape(:)
+        !! Shape of the input layer
+    end subroutine init
+
+    !module subroutine set_state(self, state)
+    !  type(rnn_layer), intent(inout) :: self
+    !  real, intent(in), optional :: state(:)
+    !end subroutine set_state
+
```
jvdp1, line 134:
```suggestion
```

src/nf/nf_rnn_layer_submodule.f90
```diff
@@ -0,0 +1,176 @@
...
+    class(rnn_layer), intent(in out) :: self
+    real, intent(in) :: input(:)
+    real, intent(in) :: gradient(:)
+    real :: db(self % output_size)
+    real :: dw(self % input_size, self % output_size)
+
+    db = gradient * self % activation % eval_prime(self % z)
+    dw = matmul(reshape(input, [size(input), 1]), reshape(db, [1, size(db)]))
+    self % gradient = matmul(self % weights, db)
+    self % dw = self % dw + dw
```
jvdp1, line 34: I recently modified these lines in [nf_dense_layer_submodule.f90](https://github.com/modern-fortran/neural-fortran/blob/118f795dbe2850032cc29623b735a8da37eac661/src/nf/nf_dense_layer_submodule.f90#L32) for better performances. The same logic could be done here IMO.

src/nf/nf_rnn_layer_submodule.f90
```diff
@@ -0,0 +1,176 @@
...
+                 + self % output_size
+
+  end function get_num_params
+
+
+  pure module function get_params(self) result(params)
+    class(rnn_layer), intent(in) :: self
+    real, allocatable :: params(:)
+
+    params = [ &
```
jvdp1, line 69: The `pack` can be avoided here, by using pointers or because it is not needed. See [here](https://github.com/modern-fortran/neural-fortran/blob/118f795dbe2850032cc29623b735a8da37eac661/src/nf/nf_dense_layer_submodule.f90#L68) for some changes.

jvdp1: Same comment for the subroutines `get_gradients` and `set_params` below.

jvdp1 (commented):
As I am currently working with `neural-fortran`, I did a quick review of this PR and left a few minor comments. Overall LGTM. Thank you.

milancurcic:
@castelao Thanks for all the additions. Sounds good.

* It may be helpful to merge `main` into here, which would reflect the few recent changes in other layers that @jvdp1 mentioned.
* Regarding loading recurrent layers from Keras HDF5 files; unless you need it, I'd say it's not a priority. If you need it, let's add it in a later PR, as there would be considerable additions needed to make it work (e.g. a Keras script example in neural-fortran/keras-snippets). That format is also no longer the latest Keras saved model format (they change too often!). Related but not part of this PR, I'd like to find an easy way to make this part (and HDF5 dependency) optional at build time. Probably via preprocessor flags in the code, but I'd need to read up how to do this in fpm.toml.
* But a standalone example program in examples/ is important.

castelao:
@jvdp1 , thanks for your suggestions. I'll work on that.

@milancurcic , yes, I have already rebased it with `main`. I'm interested in the loading from Keras output, but I see your point. I'll leave that loading capability for another PR, but I'll work on the example for this one. Thanks!

PR #172, opened 2024-03-21
Add instructions using Conda

certik:


milancurcic:
Sorry I missed this notification! I'm not a Conda user so can't test but I trust that these work. Thanks!

PR #175, opened 2024-04-16
Addition of the Loss derived type and of the MSE loss function

jvdp1:
As discussed with @milancurcic in #173 :
* addition of a derived type for loss functions
* addition of the Mean Square Error loss function

TODO:
* [x] Addition of docs
* [x] Addition of tests

### DIFF
--- a/src/nf/nf_loss_submodule.f90
+++ b/src/nf/nf_loss_submodule.f90
@@ -4,12 +4,12 @@ submodule(nf_loss) nf_loss_submodule
 
 contains
 
-  pure module function quadratic(true, predicted) result(res)
+  pure module function quadratic_eval(true, predicted) result(res)
     real, intent(in) :: true(:)
     real, intent(in) :: predicted(:)
-    real :: res(size(true))
-    res = (predicted - true)**2 / 2
-  end function quadratic
+    real :: res
+    res = sum((predicted - true)**2) / 2
+  end function quadratic_eval
 
   pure module function quadratic_derivative(true, predicted) result(res)
     real, intent(in) :: true(:)
--- a/src/nf/nf_loss_submodule.f90
+++ b/src/nf/nf_loss_submodule.f90
@@ -18,4 +18,18 @@ contains
     res = predicted - true
   end function quadratic_derivative
 
+  pure module function mse_eval(true, predicted) result(res)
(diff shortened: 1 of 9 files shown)


jvdp1:
@milancurcic what is the strategy for the tests?

src/nf/nf_loss_submodule.f90
```diff
@@ -18,4 +18,18 @@ pure module function quadratic_derivative(true, predicted) result(res)
...
+    real :: res
+    res = sum((predicted - true)**2) / size(true)
+  end function mse_eval
+
+  pure module function mse_derivative(true, predicted) result(res)
+    real, intent(in) :: true(:)
+    real, intent(in) :: predicted(:)
+    real :: res(size(true))
+    res = 2 * (predicted - true) / size(true)
+  end function mse_derivative
```
jvdp1, line 33: This function should be checked if it is valid.

milancurcic:
Thanks, @jvdp1, I'll start a test program.

milancurcic:
@jvdp1 I put a few very minimal tests that check the expected values given simple inputs. Feel free to add if you can think of better tests. I've been also thinking about how we can test for the integration of these loss functions with the network; perhaps also using simple inputs and known outputs, but pass them through the network type.

milancurcic (approved):

jvdp1:
> Feel free to add if you can think of better tests.

Thank you. These tests LGTM.

> perhaps also using simple inputs and known outputs, but pass them through the network type.
It could be a possibility. But I guess this will be more to test their support in the implementation than the functions themself. If so, would such tests be more appropriate in e.g., `test_dense_network.f90`?

milancurcic:
On second thought, let's wait on testing the integration with the network (regardless of where those tests would be defined). As we implemented general mechanisms to specify and use losses and optimizers, it's become apparent to me the important to separate model creation (i.e. via the `network_from_layers` constructor) from the "compilation", as it's done in more mature Python frameworks (e.g. in Keras you create the model first by specifying the architecture, and then in a separate step "compile" the model by passing it the loss function, the optimizer, and the eval metrics to use; this allows for example reusing the same network instance with different optimizers/losses, etc.).

I'll merge this and open a separate issue. Thank you for the PR!

PR #180, opened 2024-04-21
Remove some unused variables and a broadcast

jvdp1:


src/nf/nf_network_submodule.f90
```diff
@@ -600,11 +599,9 @@ module subroutine train(self, input_data, output_data, batch_size, &
         ! Pull a random mini-batch from the dataset
         call random_number(pos)
         batch_start = int(pos * (dataset_size - batch_size + 1)) + 1
-        batch_end = batch_start + batch_size - 1
 
         ! FIXME shuffle in a way that doesn't require co_broadcast
         call co_broadcast(batch_start, 1)
```
jvdp1, line 604: This could be called outside the loop by creating an array `batch_start(:)` and by `co_broadcast`ing  it only once

milancurcic (approved):
Thank you!

Issue #179, opened 2024-04-20
Support of a method `network % evaluate`


jvdp1:
I propose to introduce the method `evaluate` that will returns the loss value for the trained model:

```fortran
function evaluate(input_data, output_data, loss) result(res)
  real, intent(in) :: input_data(:,:)
  real, intent(in) :: output_data(:,:)
  class(loss_type), intent(in), optional :: loss
  real :: res(size(output_data, 1))
end function

**Comment by milancurcic** [comment]
I agree this would be useful. Python frameworks distinguish between losses and metrics, e.g. see https://keras.io/api/metrics/. It could be a good idea that we differentiate here as well, if we envision providing a broad range of losses and metrics. Losses can be used as metrics, but not all metrics functions are differentiable. How to best implement this without duplicating too much code from losses? The simplest approach could be to simple have a completely new `nf_metrics.f90` with `nf_metrics` module that follows the same model as `nf_loss` and its abstract derived type and function implementations, but without derivatives.

**Comment by milancurcic** [comment]
In this case, I would make the metric/loss function non-optional, for better readability of the client code.

**Comment by jvdp1** [comment]
Would somenthing like this work:

```fortran


  type, abstract :: metrics_type
  contains
    procedure(loss_interface), nopass, deferred :: eval
  end type loss_type


  type, extends(metrics_type), abstract :: loss_type
  contains
    procedure(loss_derivative_interface), nopass, deferred :: derivative
  end type loss_type
```

witth following API:
```fortran
subroutine evaluate(self, input, output, loss, metrics) result(val)
  class(network), intent(inout) :: self
  real, intent(in) :: input(:)
  real, intent(in) :: output(:)
  class(loss_type), intent(in), optional :: loss
  class(metric_type), intent(in), optional :: metrics(:) !this would also accept loss types, and therefore we don't need to duplicate them.
end subroutine
```

milancurcic:
Sounds good, in that case, is it not just:

```fortran
subroutine evaluate(self, input, output, loss, metrics) result(val)
  class(network), intent(inout) :: self
  real, intent(in) :: input(:)
  real, intent(in) :: output(:)
  class(metric_type), intent(in), optional :: metrics(:) !this would also accept loss types, and therefore we don't need to duplicate them.
end subroutine
```

milancurcic:
As discussed, the scalar variant would be

```fortran
pure real function evaluate(self, input, output, metric) result(val)
  class(network), intent(in) :: self
[...]

PR #182, opened 2024-05-06
Proposition of API for the method `network % evaluate`

jvdp1:
Related to #179

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ add_library(neural-fortran
   src/nf/nf_loss_submodule.f90
   src/nf/nf_maxpool2d_layer.f90
   src/nf/nf_maxpool2d_layer_submodule.f90
+  src/nf/nf_metrics.f90
   src/nf/nf_network.f90
   src/nf/nf_network_submodule.f90
   src/nf/nf_optimizers.f90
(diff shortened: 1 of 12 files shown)


milancurcic:
I added example use of this to examples/dense_mnist.f90.

One challenge I found is that if we use the multiple metrics variant, we can't mix and match functions defined in `nf_metrics` and those in `nf_loss` (e.g. `metrics=[mse(), corr()]` is not allowed because an array constructor must have all types the same).

I think there may be a workaround by first staging metrics as `class(metric_type), allocatable :: metrics(:)` but I couldn't quite figured it out.

Even if it can't be done, it's a minor limitation IMO.

I'll add some tests too.

milancurcic:
@jvdp1 actually, my challenge I mentioned above extends to even passing multiple metrics of the same parent type. Can you show me one example of invoking `net % evaluate_batch_1d_metrics()`?

milancurcic:
I think once we add one example of passing multiple metrics (say, in a test program), we can merge this PR.

jvdp1:
> One challenge I found is that if we use the multiple metrics variant, we can't mix and match functions defined in `nf_metrics` and those in `nf_loss` (e.g. `metrics=[mse(), corr()]` is not allowed because an array constructor must have all types the same).

Indeed, you are right. The same problem happens also with multiple metrics of the same parent type. So, it seems that `evaluate_batch_1d_metrics` is not possible and should be removed.

jvdp1:
A possible solution could be [this one ](https://gitlab.com/everythingfunctional/heterogeneous_list) proposed by Brad. But it sounds to me a bit too complex for what we want to achieve.

milancurcic:
Thank you!

PR #187, opened 2024-08-21
Make HDF5 optional under CMake

### LINKED ISSUE #186: Make HDF5 optional
It's a tedious dependency to build with across compilers and is only used for input from Keras models which not many users need. Put HDF5 code behind CPP macros and make it optional at build time.

milancurcic:
* [x] HDF5 optional with CMake using `-DUSE_KERAS_HDF5=1`
* [ ] HDF5 optional with fpm

h5fortran, or something related to it also now seems broken when build with CMake, but not with fpm. This is unrelated to this PR and can be reproduced from the main branch as well.

Closes #186


castelao:
Thank you @milancurcic !

milancurcic:
Closing in favor of a different approach that was merged in #188.

PR #178, opened 2024-04-19
Addidtion of Findneural-fortran.cmake

jvdp1:
Fixes #166 

I renamed the library from `neural` to `neural-fortran` to avoid confusion.

Issue #166, opened 2023-11-30
Building client code with CMake


aminiussi:
Hi, 

I would like to link against neural-form from my code built with CMake.

Is there a FindNeuralFortran.cmake or something similar somewhere ?

Thanks

milancurcic:
Hi Alain, I don't think so. My own knowledge of CMake is minimal (below survival level) and I've been mostly getting help from other contributors with the existing CMake setup. If we made a FindNeuralFortran.cmake, would that be expected to sit in this repo or somewhere else?

jvdp1:
based on the `Findfortran_stdlib.cmake`, I wrote the following Findneural-fortran.cmake
```cmake
# SPDX-Identifier: MIT

#[[.rst:
Find neural-fortran
-------------------

Makes the Fortran standard library (stdlib) project available.

Imported Targets
^^^^^^^^^^^^^^^^

This module provides the following imported target, if found:

``neural-fortran::neural-fortran``


Result Variables
^^^^^^^^^^^^^^^^

This module will define the following variables:

``NEURAL-FORTRAN_FOUND``
  True if the neural-fortran library is available

``NEURAL-FORTRAN_SOURCE_DIR``
  Path to the source directory of the neural-fortran library project,
  only set if the project is included as source.

``NEURAL-FORTRAN_BINARY_DIR``
  Path to the binary directory of the neural-fortran library project,
  only set if the project is included as source.

Cache variables
^^^^^^^^^^^^^^^

The following cache variables may be set to influence the library detection:

``NEURAL-FORTRAN_FIND_METHOD``
  Methods to find or make the project available. Available methods are
  - ``cmake``: Try to find via CMake config file
  - ``pkgconf``: Try to find via pkg-config file
  - ``subproject``: Use source in subprojects directory
  - ``fetch``: Fetch the source from upstream

``NEURAL-FORTRAN_DIR``
  Used for searching the CMake config file

``NEURAL-FORTRAN_SUBPROJECT``
Directory to find the neural-fortran library subproject, relative to the project root

#]]

set(_lib "neural")
set(_pkg "NEURAL-FORTRAN")
set(_url "https://github.com/modern-fortran/neural-fortran.git")

if(NOT DEFINED "${_pkg}_FIND_METHOD")
  if(DEFINED "${PROJECT_NAME}-dependency-method")
    set("${_pkg}_FIND_METHOD" "${${PROJECT_NAME}-dependency-method}")
  else()
[...]

PR #192, opened 2024-09-10
Update for cmake use of neural-fortran

mathomp4:
I recently tried building `neural-fortran` and then wanted to try and use it with an external code (which were essentially the examples herein).

In doing so, I couldn't quite get the current `Findneural-fortran.cmake` file from #178 (via #166) to work for me.

So, I did some hacking around and came up with this PR. The changes are:

1. Simplify the `Findneural-fortran.cmake` file. The original was a bit above me so I made something a bit easier to use. For this one, a user needs to set `neural-fortran_ROOT_DIR` such that `neural-fortran_ROOT_DIR/lib/libneural-fortran.a` exists. It also looks for the `.mod` files as those are also needed for external use.
2. Install the `.mod` files to `CMAKE_INSTALL_PREFIX/include` so that they can be found by the `find_package`
3. Install the `Findneural-fortran.cmake` in `CMAKE_INSTALL_PREFIX/include/cmake/neural-fortran`...which might not be the best place. I just wanted to put it somewhere in the install
4. Update the README with a bit more info

In my testing, this all seems to work both for external and internal use. For example:
```bash
cmake -B build -S . --install-prefix=$(pwd)/install -Dneural-fortran_BUILD_TESTING=YES -Dneural-fortran_BUILD_EXAMPLES=YES
cmake --build build
cmake --install build
ctest --test-dir build
```
all worked and the examples seem to as well.

I'm going to keep this draft (for now) until @milancurcic can test. I'd also like to make sure I haven't broken the workflow of @jvdp1 who added the first `Findneural-fortran.cmake` file. And I guess I'll ping @aminiussi who is the filer of #166

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: CMake Configure
-      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Release -DSERIAL=1
+      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Release
     - name: Compile
       run: cmake --build build
     - name: Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,20 +23,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - name: CMake Configure
+      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Debug -DSERIAL=1
     - name: Compile
-      run: cmake -DCMAKE_BUILD_TYPE=Debug -DSERIAL=1 . && make
+      run: cmake --build build
     - name: Test
-      run: make test
+      run: ctest --test-dir build
 
   gnu-cmake-release:
     name: gnu-cmake-release
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - name: CMake Configure
+      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Release -DSERIAL=1
     - name: Compile
-      run: cmake -DCMAKE_BUILD_TYPE=Release -DSERIAL=1 . && make
+      run: cmake --build build
     - name: Test
-      run: make test
+      run: ctest --test-dir build
 
   gnu-fpm-debug:
     name: gnu-fpm-debug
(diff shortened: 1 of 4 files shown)


mathomp4:
Huh. I didn't seem to trigger the CI...

mathomp4:
Also: I don't know how to use `fpm` so I hope I didn't break that!

milancurcic:
I think the CI is configured to only trigger CI on source changes. We should fix that!

Thanks for this, will test. fpm doesn't care about this part.

mathomp4:
> I think the CI is configured to only trigger CI on source changes. We should fix that!
> 
> Thanks for this, will test. fpm doesn't care about this part.

Ahh. I could update that in my PR? We can add `CMakeLists.txt` to the "if changed" list.

mathomp4:
Okay, `CMakeLists.txt` added and I fiddled a bit with the CI. Looks like you have to approve the CI run 😄

.github/workflows/ci.yml
```diff
@@ -21,20 +23,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - name: CMake Configure
+      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Debug -DSERIAL=1
```
milancurcic, line 27: `-DSERIAL` no longer used..

```suggestion
      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Debug
```

.github/workflows/ci.yml
```diff
@@ -21,20 +23,24 @@ jobs:
...
-      run: make test
+      run: ctest --test-dir build
 
   gnu-cmake-release:
     name: gnu-cmake-release
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - name: CMake Configure
+      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Release -DSERIAL=1
```
milancurcic, line 39: `-DSERIAL` no longer used

```suggestion
      run: cmake -B build -S . -DCMAKE_BUILD_TYPE=Release
```

milancurcic (approved):
Thank you!

milancurcic:
Thanks @mathomp4, I'm not merging in case you're still working on it, let me know when good to merge.

mathomp4:
> Thanks @mathomp4, I'm not merging in case you're still working on it, let me know when good to merge.

@milancurcic I think I'm good. Once it's in, I'll give it a test but this works for me!

mathomp4:
Thanks for merging! I'll work with `main` until next release :)

PR #194, opened 2025-01-23
Dropout layer

### LINKED ISSUE #170: Implement dropout layer
References: 

* [Keras](https://keras.io/api/layers/regularization_layers/dropout/)
* [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.Dropout.html)
* [ATHENA](https://github.com/nedtaylor/athena/blob/main/src/lib/mod_dropout_layer.f90)

milancurcic:
* [x] Dropout layer type and methods
* [x] Enable forward pass
* [x] Enable backward pass
* [x] Test
* [x] Example (in `examples/dense_mnist.f90`)

Closes #170.


ricor07:
Hello Milan. About one out of 25 times, when running test_dropout_layer this occurs: 
image

If you can't fix it today, I'll work on it tomorrow. Have a good afternoon

milancurcic:
You're right, currently the the effective dropout rate is not exactly the specified `dropout_rate` because in this (naive) implementation we simply initialize the mask with a uniform random distribution and set it to 0 if it falls below `dropout_rate` and 1 otherwise. One way to handle it is to rescale the sample from `random_number` such that the `dropout_rate` becomes the percentile of the sample.

Regarding the test failing at 1/25, notice that each test run makes 10000 dropout runs, so the effective failure rate is (10^-5). I think it's the precision issue (i.e. tolerance too small in the test), rather than the scaling of the outputs. However, you may still be correct that there's an issue with scaling the outputs, as you suggested in #195, I just don't think it's related to the test failure.

milancurcic:
Hi, I just looked at my code (haven't looked earlier when I wrote here), and I see the very serious and obvious bug in calculating the scale, which your approach fixes. No proof needed :). I'll write in the other PR on the process on opening a PR against this one, rather than a new one.

ricor07:
Hello, aren't you still merging? In the next days I'll be working on the issue i told you the other time (sorry, to look the number is boring). Which neural-fortran version should I use?

milancurcic:
Hi Ricardo, I applied your fix in https://github.com/modern-fortran/neural-fortran/pull/194/commits/59cc7e1a98434ca3a55236734cf399c6fb9ecda6.

This PR is getting closer to complete but still WIP. I need to figure out how to best enable/disable the dropout in training/inference modes, but without changing the state on the layer under the hood, which is not very user-friendly. Stay tuned.

milancurcic:
There is an outstanding bug; `dropout(0.)` should have no effect, however it does.

milancurcic:
@jvdp1 I believe this is now done; we can review together tomorrow.

src/nf/nf_dropout_layer_submodule.f90
```diff
@@ -0,0 +1,78 @@
...
+      ! not an integer, the actual dropout rate will be slightly lower.
+      self % mask = 1
+      self % mask(:int(size(self % mask) * self % dropout_rate)) = 0
+      call shuffle(self % mask)
+
+      ! Scale factor to preserve the input sum
+      self % scale = 1 / (1 - self % dropout_rate)
+
+      ! Apply dropout mask
+      self % output = input * self % mask * self % scale
```
jvdp1, line 53: I wonder what would be the most efficient. For example, if `mask` is `logical` the following could be done:
```fortran
self % output = merge(input * self%scale, 0., self%mask)
```

jvdp1 (commented):
LGTM. Here are a few cooments. I also opened a PR [here](https://github.com/milancurcic/neural-fortran/pull/3)

src/nf/nf_dropout_layer_submodule.f90
milancurcic: Good idea, it's not obvious to me either, I'll do a few tests to see if there's any difference.

milancurcic: In some very simple timing tests, using `integer` mask was indistinguishable from `real`. Using `merge` with `logical` mask appears a few percent slower based on ~10 repeats. Will keep it as is.

milancurcic:
Thank you all! Will merge.

PR #198, opened 2025-02-06
Add Input2d layer

milancurcic:
Adds `input2d` layer.

This also changes how input layers of various ranks are constructed.

Currently Fortran generic name can't distinguish between `input([10, 10])` and `input([10, 10, 10])` (1-d array as input), but it can distinguish between `input(10, 10)` and `input(10, 10, 10)` (multiple scalars as input).

Before:

```fortran
input(32)  ! input1d layer
input([3, 32, 32]) ! input3d layer
```

This PR:

```fortran
input(32) ! input1d layer
input(32, 32) ! input2d layer
input(3, 32, 32) ! input3d layer
```

@oneadder please review.

This PR is in support of removing the crutch in #197.

TODO:

* [x] `network % {forward,predict,predict_batch}_2d`
* [x] `input2d` tests
* [x] `input2d` example
* [x] Update CMake
* [x] Update README.md

PR #199, opened 2025-02-09
Multihead attention

OneAdder:
Hello, Milan! I hope, I'm not bothering you too much with my pull requests, but this is a good one. At this stage it is a draft of MultiHead Attention. It cannot be merged until work on [`input2d_layer`](https://github.com/modern-fortran/neural-fortran/pull/198) and [`linear2d_layer`](https://github.com/modern-fortran/neural-fortran/pull/197) is completed.
Implementation of [`dropout`](https://github.com/modern-fortran/neural-fortran/pull/194) would also help improve MHA, but it can be added later.

## MultiHead Attention

MultiHead Attention is the main component of [Transformer](https://proceedings.neurips.cc/paper_files/paper/2017/file/3f5ee243547dee91fbd053c1c4a845aa-Paper.pdf) architecture, which is the most advanced modern approach in the area of Natural Language Processing, as well as some other areas.
Here I propose an implementation based on the [Transformer](https://proceedings.neurips.cc/paper_files/paper/2017/file/3f5ee243547dee91fbd053c1c4a845aa-Paper.pdf) article. It works and its output conforms with SOTA implementation in PyTorch.

## Python Reference
https://github.com/OneAdder/neural-fortran-references/blob/main/self_attention.py


ricor07:
Hello Micheal, I saw your pull requests and I think what you do is very interesting. Could you take a look at mine? https://github.com/milancurcic/neural-fortran/pull/2

What you have to look here is not the locally connected 1d layer but rather the reshape architecture I am trying to make. Your help would be very appreciated. Thanks

milancurcic:
Amazing, thank you! Yes, let's wrap up Input2d in #198, then Linear2d in #197, to avoid the 3d crutch.

ricor07:
@OneAdder i see you are talking about a 2d handling. Would you like to make this together? I have to make this as well since I'm implementing a conv 1d layer

milancurcic:
Hi guys, thanks for pushing this forward. Today I'm finishing a lot of busywork with some proposals so next week I'm getting back more actively with neural-fortran work, and will be able to contribute to the reviews more actively

OneAdder:
@ricor07 Great idea! But we have a problem with generics again. The issues is that a `predict_2d` and `predict_1d_batch` have the same rank of input. I think we should simply make a separate generic for `predict_batch`. @milancurcic your thoughts?

ricor07:
Yes, I think we can make a generic predict. But I suggest you to create a new branch

milancurcic:
I think it's fine to make `predict_batch` its own generic name because it's getting in the way. 👍

OneAdder:
@milancurcic Done, here: https://github.com/modern-fortran/neural-fortran/pull/198
@ricor07 There is still one piece of the puzzle missing: a general flatten layer with interfaces for both 3D and 2D. Do you want to implement it? Or I can do it together with linear2d layer

ricor07:
You can make it. I'll work on maxpool

milancurcic:
Hi Michael, just in case you weren't aware, Ondrej has a procedural implementation of GPT-2, including a MHA subroutine here: https://github.com/certik/fastGPT/blob/main/gpt2.f90. You can use it as a reference Fortran implementation if you need.

When you're ready for me to start playing with this PR, just mark it Ready for review.

OneAdder:
@milancurcic The attention backward and forward have both been done and tested like a week ago. I mostly used Attention Is All You Need paper as a reference as it is pretty straightforward and includes most of the formulae. And, besides, that example is not particularly useful here as it doesn't implement back prop. But we can reuse at a later point the piece that loads weights from HuggingFace.
I'm now working on figuring out how to create an example for MHA. So, most of the code is ready for review

OneAdder:
Except one thing: redundant arguments. I'll fix that

OneAdder:
So, my example works, the full check list:
- [ ] Compile threaded version (I still didn't manage to figure out what Debian packages I need)
- [ ] Figure out how to optimize out two extra passes through `n_heads` without sacrificing readability
- [ ] Remove redundant arguments in constructors
- [ ] Add an example

@milancurcic I have this text classification example which can also be used as an example for MHA. But it's large and doesn't make much sense until I add Positional Encoding (WIP). Should I add it here anyway or should I add it a later point in a separate PR? I can come up with a small toy example here

milancurcic:
Excellent, thanks, I'll begin reviewing tomorrow morning. How large is it? In the past, I'd upload data tarballs as an attachment in a GitHub issue and use its permanent URL to download from programs. That way data storage doesn't go into git history. But there's some size limit. Our MNIST data used here is stored like that.

milancurcic:
For example of datasets stored here, see https://github.com/modern-fortran/neural-fortran/blob/c316ee1ed1c318a0f74fb63c8aec04dda2674aa7/src/nf/nf_datasets.f90#L18-L29

OneAdder:
@milancurcic I think it's ready for review! I'll add the complicated example later. At this stage I added a simple example that converges nicely and doesn't require datasets and extra deps

src/nf/nf_cross_attention_layer.f90
```diff
@@ -0,0 +1,66 @@
+module nf_cross_attention_layer
+  use iso_fortran_env, only: stderr => error_unit
+  use nf_activation, only: softmax
+  use nf_linear2d_layer, only: linear2d_layer
+  use nf_multihead_attention_layer, only: multihead_attention_layer
+
+  implicit none
+
+  type, extends(multihead_attention_layer) :: cross_attention_layer
```
OneAdder, line 9: That is intentional that there is no plumbing for this one yet. I suggest that we add it at later stage when we have more components for seq2seq models. At this stage it can be added like this: without any public access

test/test_multihead_attention_layer.f90
```diff
@@ -0,0 +1,412 @@
...
+  end if
+
+contains
+  function allclose(x, y) result(res)
+    real, intent(in) :: x(:)
+    real, intent(in) :: y(:)
+    logical :: res
+
+    res = all(abs(x - y) <= (1e-06 + 1e-05 * abs(y)))
+  end function allclose
```
OneAdder, line 47: Suggestion for future: create `nf_utils.f90` (or similar) and put this procedure there

src/nf/nf_multihead_attention_submodule.f90
```diff
@@ -0,0 +1,343 @@
...
+    real, intent(in) :: query(:, :, :)
+    real, intent(in) :: key(:, :, :)
+    integer :: head
+    ! create attention matrix for each sequence in each batch
+    do concurrent(head = 1: self % n_heads)
+      self % attention_matrix(:, :, head) = matmul(query(:, :, head), transpose(key(:, :, head)))
+    end do
+  end subroutine create_attention_matrix
+
+  pure module subroutine normalize_attention_matrix(self, attention_mask)
```
OneAdder, line 177: `attention_mask` is not accessible to the users by design at this point. It will be used by transformer decoder later and I'll add corresponding logic later

src/nf/nf_multihead_attention.f90
```diff
@@ -0,0 +1,170 @@
...
+      real :: output(self % sequence_length, self % head_size, self % n_heads)
+    end function split_heads
+
+    pure module subroutine create_attention_matrix(self, query, key)
+      !! Create attention matrix for query and key
+      !! Output dimensions: sequence_length, sequence_length, n_heads
+      class(multihead_attention_layer), intent(in out) :: self
+      real, intent(in) :: query(:, :, :)
+      real, intent(in) :: key(:, :, :)
+      integer :: head
```
jvdp1, line 115:
```suggestion
```

src/nf/nf_multihead_attention.f90
```diff
@@ -0,0 +1,170 @@
...
+
+    pure module subroutine normalize_attention_matrix(self, attention_mask)
+      !! Create attention matrix for query and key
+      !! Output dims: sequence_length, sequence_length, n_heads
+      class(multihead_attention_layer), intent(in out) :: self
+      !! (sequence_length, sequence_length, n_heads)
+      real, optional, intent(in) :: attention_mask(:, :, :)
+      !! (sequence_length, sequence_length, n_heads)
+      real, allocatable :: output(:, :, :)
+      integer :: head, seq
```
jvdp1, line 126:
```suggestion
```

src/nf/nf_multihead_attention.f90
```diff
@@ -0,0 +1,170 @@
...
+    end subroutine normalize_attention_matrix
+
+    pure module subroutine scaled_dot_product_attention(self, value)
+      !! Create scaled dot product attention
+      !! Output dims: sequence_length, head_size, n_heads
+      class(multihead_attention_layer), intent(in out) :: self
+      real, intent(in) :: value(:, :, :)
+      integer :: head
```
jvdp1, line 134:
```suggestion
```

jvdp1 (commented):
@OneAdder Nice set of PRs. I added a few suggestions. Feel free to ignore them.
Furthermore, I usually try to avoid `reshape` and using pointers instead, mainly for performance reasons. I think this approach could be used at some places, but it could be for a next PR.

milancurcic:
@OneAdder:

`self_attention_layer` and `cross_attention_layer` both extend `multihead_attention_layer`, but only `self_attention` is exposed in the user-facing module `nf` and in constructors.

1. Should `cross_attention` be made accessible to the user like `self_attention` is?
2. Is `multihead_attention` intended to be used only internally as a building block, or is there value to making it part of the public API?

The answers to these questions will also guide which layers to include in the layers table in the README.

OneAdder:
@jvdp1 Great idea! I think we should extend it even further and make smth like `nf_utils` that does those inplace reshapes. So that we could use it whenever possible. But I think this PR is already pretty long. So, I suggest we do it later

OneAdder:
@milancurcic 
1. Yes, but at this point we don't have any layers to make it useful. Therefore I made it ready to be included but not actually included as of yet. I suggest it makes it to public API later.
2. I don't know any scenarios where Attention is neither self nor cross. If there are, I think they should also be separate classes extending `multihead_attention_layer`, as it itself doesn't implement `forward` and `backward` methods

milancurcic:
Thank you!!

PR #201, opened 2025-02-13
Locally connected layer

ricor07:
Continuing https://github.com/milancurcic/neural-fortran/pull/2#event-16297832061 on the right repository. Reshape_generalized works, locally_connected_1d still does not. This is, of course, still a draft. My final goal will be making the layer work and generalizing ALL the 1d-3d function which are horrible to see.

Edit: sorry, I mistakenly closed the other PR.


OneAdder:
@ricor07 It seems, most of the problems with 2D will resolved with this: https://github.com/modern-fortran/neural-fortran/pull/198

ricor07:
Great. Tomorrow I'm going to try your implementation

ricor07:
Hello, after the recent updates I think that my reshape_generalized ceased to work. For these reasons, I preferred to switch to a reshape2d implementation. Of course, after making this work the goal would be to create a unique reshape file. The cnn mnist 1d file, with the locally_connected_1d layer, does still not work but it is normal since I have not worked on its connection to layer and network yet. Tomorrow I will. Sorry for the delay but I had some difficulties

src/nf/nf_reshape2d_layer_submodule.f90
```diff
@@ -0,0 +1,50 @@
+submodule(nf_reshape2d_layer) nf_reshape2d_layer_submodule
```
jvdp1, line 1: For efficiency, are there no ways to use pointers instead of `pack` and `reshape`? I couldn't find one.

ricor07: I think there aren't. At least, there are no ways I am fond of.

ricor07:
Hello, sorry for not having given updates but I had been busy. I almost finished. I only have to fix one bug in cnn_network and then this draft will be ready for review, i think already tomorrow.

src/nf/nf_reshape2d_layer_submodule.f90
```diff
@@ -0,0 +1,50 @@
...
+  end function reshape2d_layer_cons
+
+
+  pure module subroutine backward(self, input, gradient)
+    class(reshape2d_layer), intent(in out) :: self
+    real, intent(in) :: input(:)
+    real, intent(in) :: gradient(:,:)
+    ! The `input` dummy argument is not used but nevertheless declared
+    ! because the abstract type requires it.
+    self % gradient = pack(gradient, .true.)
```
jvdp1, line 22: What is the purpose of this?
Would the following not more eefficient?
```suggestion
        real :: g_(:) => null()
        g_(1:size(gradient)) => gradient
        self % gradient = g_
        nullify(g_)
```

src/nf/nf_reshape2d_layer_submodule.f90
```diff
@@ -0,0 +1,50 @@
...
+  end subroutine backward
+
+
+  pure module subroutine forward(self, input)
+    class(reshape2d_layer), intent(in out) :: self
+    real, intent(in) :: input(:)
+    self % output = reshape(input, self % output_shape)
```
jvdp1, line 29: A pointer could be used here, instead of `reshape` to avoid a tmp copy of `input`

jvdp1 (commented):
Overall LGTM. Here are some suggestions.

src/nf/nf_datasets_mnist_submodule.f90
```diff
@@ -50,9 +50,9 @@ module subroutine load_mnist(training_images, training_labels, &
     real, allocatable, intent(in out), optional :: testing_labels(:)
 
     integer, parameter :: dtype = 4, image_size = 784
-    integer, parameter :: num_training_images = 50000
-    integer, parameter :: num_validation_images = 10000
-    integer, parameter :: num_testing_images = 10000
+    integer, parameter :: num_training_images = 500
+    integer, parameter :: num_validation_images = 100
+    integer, parameter :: num_testing_images = 100
```
milancurcic, line 55: I assume you didn't intend to commit this, correct?

milancurcic:
Thanks @ricor07. This is a huge PR. Few things before I go in-depth: Is `conv_mnist_1d` supposed to converge? with full MNIST dataset (50k/10k/10k), it stays at a 10% accuracy (random guess). If I keep the decimated MNIST dataset from your PR (500/100/100), it goes up to about 60% before collapsing to 10%.

ricor07:
@milancurcic i wanted your opinion on this thing.

First of all, I added the test test_conv1d_network on the model of test_conv2d_network. I noticed that often the tests fail, because the values outputted in get_params (so, kernel and bias) diverge. This problem does not depend on the learning rate, nor on the layer structure. 

When the tests fail, the gradients (dw and db) take more time to converge to a certain value.

To be fair, I don't know what to do to solve this. Maybe you can think to something. Tomorrow I'll try to debug this line by line.

milancurcic:
Would it be a good solution to just reduce the convergence tolerance in the test by an order or two of magnitude to ensure convergence?

ricor07:
No, I already tried. 

While debugging, I found out that in conv2d_submodule gradient is always 0.

First line, gdz update, this is gradient * something, but gradient is initialized to 0 so gdz is 0
db is sum of gdz, so always 0
dw = dw + input * gdz, so 0
gradient = gradient + sum(gdz) * kernel, always 0.

I printed out the gradient and it is, indeed, always equal to 0.

Am I wrong or is there a huge error?
I say this because I wrote conv1d totally out from conv2d so I think this error could have been propagating.
Sorry if I make you lose time, but could you take a look?

milancurcic:
OK, awesome, I'll take a look, it may very well be another bug in conv2d!

ricor07:
Plus, where do kernel and biases get updated?

milancurcic:
It's here:

https://github.com/modern-fortran/neural-fortran/blob/ed8b3404264c0932b8c6d712f1dc29203f3d69e8/src/nf/nf_network_submodule.f90#L625-L627

ricor07:
Do I need to add conv1d under line 620 of what you posted?

src/nf/nf_conv1d_layer_submodule.f90
```diff
@@ -0,0 +1,187 @@
...
+    implicit none
+    class(conv1d_layer), intent(in out) :: self
+    real, intent(in) :: input(:,:)
+    integer :: input_channels, input_width
+    integer :: j, n
+    integer :: iws, iwe, half_window
+
+    input_channels = size(input, dim=1)
+    input_width    = size(input, dim=2)
+    half_window = self % kernel_size / 2
```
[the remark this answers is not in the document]
ricor07: Not required

example/cnn_mnist_1d.f90
```diff
@@ -0,0 +1,67 @@
...
+        net, validation_images, label_digits(validation_labels)) * 100, ' %'
+  
+    end do epochs
+  
+    print '(a,f5.2,a)', 'Testing accuracy: ', &
+      accuracy(net, testing_images, label_digits(testing_labels)) * 100, '%'
+  
+  contains
+  
+    real function accuracy(net, x, y)
```
OneAdder, line 53: Offtop thought: maybe add smth like `nf_metrics` at some point in the future?

ricor07: Going to raise an issue

src/nf/nf_activation.f90
```diff
@@ -25,12 +25,14 @@ module nf_activation
   contains
     procedure(eval_1d_i), deferred :: eval_1d
     procedure(eval_1d_i), deferred :: eval_1d_prime
+    procedure(eval_2d_i), deferred :: eval_2d
```
OneAdder, line 28: Question to @milancurcic :
Do we really need 2D and 3D activation functions? I used 1D activations for 2D shapes without any issues and overcomplication of my code.
And, if we do need them, maybe use assumed-ranks here? I'm not really a fan of those copypasted generics

src/nf/nf_conv1d_layer_submodule.f90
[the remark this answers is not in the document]
OneAdder: I can't find any usages of `half_window` variable. Am I missing something or it can be deleted?

src/nf/nf_locally_connected_1d_submodule.f90
```diff
@@ -0,0 +1,152 @@
...
+    class(locally_connected_1d_layer), intent(in out) :: self
+    real, intent(in) :: input(:,:)
+    integer :: input_channels, input_width
+    integer :: j, n
+    integer :: iws, iwe
+
+    input_channels = size(input, dim=1)
+    input_width    = size(input, dim=2)
+
+    do j = 1, self % width
```
OneAdder, line 65: `do concurrent`?

OneAdder: Here and below

src/nf/nf_locally_connected_1d_submodule.f90
```diff
@@ -0,0 +1,152 @@
...
+
+    dw_local = 0.0
+    self % gradient = 0.0
+
+    do n = 1, self % filters
+       do j = 1, output_width
+          iws = j
+          iwe = j + self % kernel_size - 1
+          do k = 1, self % channels
+             dw_local(n, j, k, :) = dw_local(n, j, k, :) + input(k, iws:iwe) * gdz(n, j)
```
OneAdder, line 109: Looks difficult, perhaps, a comment?

src/nf/nf_locally_connected_1d_submodule.f90
```diff
@@ -0,0 +1,152 @@
...
+  pure module function get_num_params(self) result(num_params)
+    class(locally_connected_1d_layer), intent(in) :: self
+    integer :: num_params
+    num_params = product(shape(self % kernel)) + product(shape(self % biases))
+  end function get_num_params
+
+  module function get_params(self) result(params)
+    class(locally_connected_1d_layer), intent(in), target :: self
+    real, allocatable :: params(:)
+    params = [reshape(self % kernel, [size(self % kernel)]), reshape(self % biases, [size(self % biases)])]
```
OneAdder, line 128: Maybe I'm missing something, can it not be reduced to:
```fortran
params = [self % kernel, self % biases]
```
?

src/nf/nf_reshape2d_layer.f90
```diff
@@ -0,0 +1,77 @@
+module nf_reshape2d_layer
```
OneAdder, line 1: Thank you!!

OneAdder:
@ricor07 Great job! I really like it!

src/nf/nf_locally_connected_1d_submodule.f90
ricor07: I don't think it's needed. It computes dw just as in the conv layers but with another dimension added.

ricor07:
@OneAdder thank you for your review, I did the adjustments you suggested.

ricor07:
I tried to solve conflicts with main but i think i have to merge more accurately. Is that needed, since you only have to merge the branch?

milancurcic:
If there were conflicts of this branch with main, then they had to be reconciled by merging main into here. There don't seem to be merge conflict now.

ricor07:
Ok, solved

milancurcic:
No, `main` is protected so it can only be changed by merging a PR. Now we have to resolve conflicts in this branch. There are a few very small conflicts here: https://github.com/modern-fortran/neural-fortran/pull/201/conflicts. You can resolve them by hand, or I can resolve them later, it's not a big issue.

ricor07:
Choose the layout you prefer. Thank you

example/cnn_mnist_1d.f90
```diff
@@ -0,0 +1,67 @@
...
+    integer, parameter :: num_epochs = 250
+  
+    call load_mnist(training_images, training_labels, &
+                    validation_images, validation_labels, &
+                    testing_images, testing_labels)
+  
+    net = network([ &
+      input(784), &
+      reshape2d([28,28]), &
+      locally_connected_1d(filters=8, kernel_size=3, activation=relu()), &
```
jvdp1, line 24: Why is a `reshape2d` needed if `locally_connected_1d` is used?

src/nf/nf_conv1d_layer_submodule.f90
```diff
@@ -0,0 +1,183 @@
...
+
+  use nf_activation, only: activation_function
+  use nf_random, only: random_normal
+
+  implicit none
+
+contains
+
+  module function conv1d_layer_cons(filters, kernel_size, activation) result(res)
+    implicit none
```
jvdp1, line 11:
```suggestion
```
These `implicit none` are not needed, as already covered by the `implicit none` of the submodule.

src/nf/nf_conv1d_layer_submodule.f90
```diff
@@ -0,0 +1,183 @@
...
+
+    !--- Compute the local gradient gdz = (dL/dy) * sigma'(z) for each output.
+    do j = 1, output_width
+       gdz(:, j) = gradient(:, j) * self % activation % eval_prime(self % z(:, j))
+    end do
+
+    !--- Compute bias gradients: db(n) = sum_j gdz(n, j)
+    do n = 1, self % filters
+       db_local(n) = sum(gdz(n, :), dim=1)
+    end do
```
jvdp1, line 118: The loop is not needed here IMO
```suggestion
       db_local(:) = sum(gdz(:, :), dim=2)
```
It should be checked if it should be `dim=1` or `dim=2`

src/nf/nf_conv1d_layer_submodule.f90
```diff
@@ -0,0 +1,183 @@
...
+    real, intent(in) :: params(:)
+
+    if (size(params) /= self % get_num_params()) then
+      error stop 'conv1d_layer % set_params: Number of parameters does not match'
+    end if
+
+    self % kernel = reshape(params(:product(shape(self % kernel))), shape(self % kernel))
+    associate(n => product(shape(self % kernel)))
+      self % biases = params(n + 1 : n + self % filters)
+    end associate
```
jvdp1, line 179:
```suggestion
    associate(n => product(shape(self % kernel)))
        self % kernel = reshape(params(:n), shape(self % kernel))
        self % biases = params(n + 1 : n + self % filters)
    end associate
```

example/cnn_mnist_1d.f90
jvdp1: Ok. I guess that the `reshape2d` has a similar function to the `input_shape` in `keras`

milancurcic:
Thank you for this PR, it's a great addition! I'll take a day or two more to review in more detail as it's so big. I'd also like to look whether it's feasible to hide conv1d/2d, maxpool1d/2d, and reshape2d/3d behind generic layer constructors `conv`, `maxpool`, and `reshape`, since the specific implementations depending on the input rank seem to be an internal detail. Do you think it would be a nicer UI to have generic names?

ricor07:
Yes, I do. In a few days I'll start implementing other stuff like averagepooling and i will also group those under generic names. 

If you have time, I think i made a mistake with naming locally_connected_layer. I think that 1d does not need an underline before it. Could you fix this? Thanks

src/nf/nf_activation.f90
milancurcic: I'll look into it more but I think what you noticed that you could do is a non-standard behavior provided by the compiler, and in that case, not portable.

milancurcic (approved):
Thank you!

PR #202, opened 2025-02-16
Generic flatten (2d and 3d)

milancurcic:
I attempted to make a generic `flatten` so that the user doesn't need to do `flatten2d`. It seems like it will work.

In support of `Linear2d` (#197)

### DIFF
--- a/src/nf/nf_flatten_layer.f90
+++ b/src/nf/nf_flatten_layer.f90
@@ -18,13 +18,20 @@ module nf_flatten_layer
     integer, allocatable :: input_shape(:)
     integer :: output_size
 
-    real, allocatable :: gradient(:,:,:)
+    real, allocatable :: gradient_2d(:,:)
+    real, allocatable :: gradient_3d(:,:,:)
     real, allocatable :: output(:)
 
   contains
 
-    procedure :: backward
-    procedure :: forward
+    procedure :: backward_2d
+    procedure :: backward_3d
+    generic :: backward => backward_2d, backward_3d
+
+    procedure :: forward_2d
+    procedure :: forward_3d
+    generic :: forward => forward_2d, forward_3d
+
     procedure :: init
 
   end type flatten_layer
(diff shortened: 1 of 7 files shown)


ricor07:
LGTM

ricor07 (approved):

src/nf/nf_flatten_layer.f90
```diff
@@ -18,13 +18,20 @@ module nf_flatten_layer
     integer, allocatable :: input_shape(:)
     integer :: output_size
 
-    real, allocatable :: gradient(:,:,:)
+    real, allocatable :: gradient_2d(:,:)
+    real, allocatable :: gradient_3d(:,:,:)
```
OneAdder, line 22: Yep, I thought about that but decided not to make the code even less SOLID

src/nf/nf_flatten_layer.f90
```diff
@@ -18,13 +18,20 @@ module nf_flatten_layer
...
 
-    procedure :: backward
-    procedure :: forward
+    procedure :: backward_2d
+    procedure :: backward_3d
+    generic :: backward => backward_2d, backward_3d
+
+    procedure :: forward_2d
+    procedure :: forward_3d
+    generic :: forward => forward_2d, forward_3d
```
OneAdder, line 33: Perhaps, just make it one function with assumed-rank input?
```fortran
  pure module subroutine forward(self, input)
    class(flatten_layer), intent(in out) :: self
    real, intent(in) :: input(..)
    select rank(input)
      rank(2)
        self % output = pack(input, .true.)
      rank(3)
        self % output = pack(input, .true.)
      rank default
        error stop "Unsupported rank of input"
    end select
  end subroutine forward
```
It will reduce boilerplate a little

OneAdder:
Thank you! I'll rebase and test it out!

src/nf/nf_flatten_layer.f90
OneAdder: But here we have a choice between SOLID and less boilerplate, I think I agree that the second one is better

milancurcic: Yes, and, most importantly for me, this approach allows for a unified API (only one `flatten()` for the user).

src/nf/nf_flatten_layer.f90
milancurcic: Cool! If it works, let's do it.

OneAdder: done

OneAdder (approved):
Everything works. Great job!

Issue #145, opened 2023-06-22
CNN training on MNIST does not converge


milancurcic:
* `cnn_mnist` example which trains a CNN network on MNIST data stays at random (10%) accuracy over epochs;
* `cnn_from_keras` example which loads a pre-trained CNN from Keras and achieves expected high accuracy (90.14%)

The above suggests that the forward passes of `conv2d`, `maxpool2d`, and `flatten` layers are implemented correctly.

The culprit may be in the implementation of `backward` methods for any of these layers, or in the backward flow of data.

This should be fixed before the release of v0.13.0.

certik:
Here is an example output that I am getting:
```console
$ ./build/gfortran_095607A870B6C445/example/cnn_mnist 
Layer: input
------------------------------------------------------------
Output shape: 784
Parameters: 0

Layer: reshape
------------------------------------------------------------
Input shape: 784
Output shape: 1 28 28
Parameters: 0
Activation: 

Layer: conv2d
------------------------------------------------------------
Input shape: 1 28 28
Output shape: 8 26 26
Parameters: 80
Activation: relu

Layer: maxpool2d
------------------------------------------------------------
Input shape: 8 26 26
Output shape: 8 13 13
Parameters: 0
Activation: 

Layer: conv2d
------------------------------------------------------------
Input shape: 8 13 13
Output shape: 16 11 11
Parameters: 1168
Activation: relu

Layer: maxpool2d
------------------------------------------------------------
Input shape: 16 11 11
Output shape: 16 5 5
Parameters: 0
Activation: 

Layer: flatten
------------------------------------------------------------
Input shape: 16 5 5
Output shape: 400
Parameters: 0
Activation: 

Layer: dense
------------------------------------------------------------
Input shape: 400
Output shape: 10
Parameters: 4010
Activation: softmax

Epoch  1 done, Accuracy:  9.91 %
Epoch  2 done, Accuracy:  9.91 %
Epoch  3 done, Accuracy:  9.91 %
Epoch  4 done, Accuracy:  9.91 %
...
```
It will stay at this percentage.

certik:
Git bisect reveals:
```
6bbc28d123cdec20140331edc60df106d518a202 is the first bad commit
commit 6bbc28d123cdec20140331edc60df106d518a202
Author: Milan Curcic <caomaco@gmail.com>
Date:   Thu Jun 22 11:27:03 2023 -0400

[...]

PR #203, opened 2025-02-17
Layernorm

OneAdder:
# Layer Normalization

Layer Normalization applies normalization over a sequence of inputs by the following formula:

$\frac{x - \bar{x}}{\sqrt{var(x) + \epsilon}} \cdot \gamma + \beta$

Where:
* $\bar{x}$ is arithmetic mean over the last dimension
* $var(x)$ is variance over the last dimension
* $\epsilon$ is a scaling constant (small)
* $\gamma$ and $\beta$ are trainable parameters

### Reference

Lei Ba, J., Kiros, J. R., & Hinton, G. E. (2016). Layer normalization. ArXiv e-prints, [arXiv-1607](https://arxiv.org/pdf/1607.06450).

### Status

Ready


src/nf/nf_layernorm_submodule.f90
```diff
@@ -0,0 +1,110 @@
...
+
+    ! normalize mu by variance by first axis
+    do concurrent(i = 1: self % model_dimension)
+      normalized(:, i) = self % mu(:, i) / self % sigma
+    end do
+
+    ! forward through trainable params gamma and beta
+    do concurrent(i = 1: self % sequence_length)
+      self % output(i, :) = normalized(i, :) * self % gamma + self % beta
+    end do
```
jvdp1, line 35: It seems that these two loops could be combined, such that the variable `normalized` would not be needed.

test/test_layernorm.f90
```diff
@@ -0,0 +1,93 @@
...
+  call layernorm % init([3, 4])
+
+  call test_layernorm_forward(layernorm, sample_input, ok)
+  call test_layernorm_backward(layernorm, sample_input, sample_gradient, ok)
+
+  if (ok) then
+    print '(a)', 'test_layernorm_layer: All tests passed.'
+  else
+    write(stderr, '(a)') 'test_layernorm_layer: One or more tests failed.'
+    stop 1
```
jvdp1, line 21:
```suggestion
    error stop 1
```

src/nf/nf_layernorm_submodule.f90
OneAdder: Yes, I'll fix that

OneAdder: Can you explain how would you do that? The only thing I can think of is making a nested loop like this:
```fortran
do concurrent(i = 1: self % sequence_length, j = 1: self % model_dimension)
```
But that'll obviously make it $n \cdot m$ complexity

jvdp1: I was indeed thinking to such a nested loop, as it is currently twice the case implicitely. but I might be wrong, and this could be checked at a future stage.

OneAdder: In this case I'd definitely use some extra memory to reduce processor complexity to $n + m$

OneAdder:
@milancurcic @jvdp1 Ready for review

src/nf/nf_layernorm_submodule.f90
```diff
@@ -0,0 +1,158 @@
...
+    res % eps = 1e-5
+  end function layernorm_layer_cons
+
+  pure module subroutine forward(self, input)
+    class(layernorm_layer), intent(in out) :: self
+    real, intent(in) :: input(:, :)
+    real, allocatable :: normalized(:, :)
+    integer :: i
+
+    allocate(normalized(self % sequence_length, self % model_dimension))
```
jvdp1, line 17:
```suggestion
    allocate(normalized, mold=self % mu)
```

jvdp1 (approved):
LGTM @OneAdder . Nice work!

src/nf/nf_layer_constructors.f90
```diff
@@ -222,15 +223,23 @@ module function linear2d(out_features) result(res)
...
+      !! Rank-2 (sequence_length, out_features) self attention constructor.
+      !! sequence_length and model_dimension are determined at layer initialization, based on the
+      !! output shape of the previous layer.
+      integer, intent(in) :: num_heads
+        !! Number of attention heads
+      type(layer) :: res
+        !! Resulting layer instance
+    end function self_attention
+
+    module function layer_normalization() result(res)
```
milancurcic, line 236: How about `layernorm` for the constructor? Shorter, and `norm` is quite ubiquitous to mean "normalization". PyTorch uses `LayerNorm` and Keras uses `LayerNormalization`.

src/nf/nf_layernorm_submodule.f90
```diff
@@ -0,0 +1,158 @@
...
+    res % eps = 1e-5
+  end function layernorm_layer_cons
+
+  pure module subroutine forward(self, input)
+    class(layernorm_layer), intent(in out) :: self
+    real, intent(in) :: input(:, :)
+    real, allocatable :: normalized(:, :)
+    integer :: i
+
+    allocate(normalized, mold=self % mu)
```
milancurcic, line 17: Since the shape of `mu` is determined at `init`, so is the shape of `normalized`.  It's inefficient to allocate+deallocate an array in every `forward` call. Let's allocate `normalized` in `init` as well.

src/nf/nf_layernorm_submodule.f90
```diff
@@ -0,0 +1,158 @@
...
+
+  pure module subroutine backward(self, input, gradient)
+    class(layernorm_layer), intent(in out) :: self
+    real, intent(in) :: input(:, :)
+    real, intent(in) :: gradient(:, :)
+    real, allocatable :: one_over_sigma(:, :)
+    real, allocatable :: gradient_by_gamma_over_sigma(:, :)
+
+    allocate(one_over_sigma(self % sequence_length, self % model_dimension))
+    allocate(gradient_by_gamma_over_sigma(self % sequence_length, self % model_dimension))
```
milancurcic, line 48: Same for these, should be allocated at `init` time

test/test_layernorm.f90
```diff
@@ -0,0 +1,151 @@
...
+  use iso_fortran_env, only: stderr => error_unit
+  use nf_layernorm_layer, only: layernorm_layer
+  use nf, only: sgd
+  implicit none
+
+  logical :: ok = .true.
+  type(layernorm_layer) :: layernorm
+  real :: sample_input(3, 4) = reshape([0.0, 10.1, 0.2, 10.3, 0.4, 10.5, 0.6, 10.7, 10.8, 0.9, 0.11, 0.12], [3, 4])
+  real :: sample_gradient(3, 4) = reshape([0.1, 3., 2., 0.1, 3., 3., 0.1, 2., 0.1, 3., 0.1, 3.], [3, 4])
+
```
milancurcic, line 11: Test the constructor too.

milancurcic:
Great, thank you! I left several comments.

test/test_layernorm.f90
```diff
@@ -0,0 +1,151 @@
...
+  layernorm = layernorm_layer()
+  call layernorm % init([3, 4])
+
+  call test_layernorm_forward(layernorm, sample_input, ok)
+  call test_layernorm_backward(layernorm, sample_input, sample_gradient, ok)
+  call test_layernorm_gradients(sample_input, sample_gradient, ok)
+
```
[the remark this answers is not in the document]
OneAdder: @milancurcic I have made such test here. The only issue here is that margin for error should be rather high (`0.1`) and a lot of training epochs. The issue here is that to test convergence we use LayerNormalization as an activation function which works but is inefficient

test/test_layernorm.f90
```diff
@@ -0,0 +1,193 @@
...
+    integer :: epochs = 10000
+
+    net = network([&
+        input(2, 3),&
+        linear2d(3),&
+        layernorm(),&
+        flatten()&
+    ])
+
+    ! Kaiming weights to achieve semblance of convergance
```
OneAdder, line 172: If we want LayerNorm used as activation function to converge, we need to at least use correct weights initialization. Otherwise, the net will not be able to consistently converge and we'll get https://github.com/modern-fortran/neural-fortran/issues/145 once more

README.md
```diff
@@ -34,8 +34,9 @@ Read the paper [here](https://arxiv.org/abs/1902.06714).
 | Dropout | `dropout` | `dense`, `flatten`, `input1d` | 1 | ✅ | ✅ |
 | Convolutional (2-d) | `conv2d` | `input3d`, `conv2d`, `maxpool2d`, `reshape` | 3 | ✅ | ✅(*) |
 | Max-pooling (2-d) | `maxpool2d` | `input3d`, `conv2d`, `maxpool2d`, `reshape` | 3 | ✅ | ✅ |
-| Linear (2-d) | `linear2d` | `input2d`, `linear2d`, `self_attention` | 2 | ✅ | ✅ |
-| Self-attention | `self_attention` | `input2d`, `linear2d`, `self_attention` | 2 | ✅ | ✅ |
+| Linear (2-d) | `linear2d` | `input2d`, `layernorm`, `linear2d`, `self_attention` | 2 | ✅ | ✅ |
+| Self-attention | `self_attention` | `input2d`, `layernorm`, `linear2d`, `self_attention` | 2 | ✅ | ✅ |
+| Layer Normalization | `layernorm` | `linear2d`, `self_attention` | 2 | ✅ | ✅ |
```
milancurcic, line 39: @OneAdder can you please check that I did this correctly?

milancurcic (approved):
Thank you!!

README.md
OneAdder: Yes, looks good!

PR #204, opened 2025-02-17
Weight initialization

OneAdder:
# Weights Initialization

Added functions for Xavier and Kaiming. The rule of thumb here:
* S-shaped activation (`tanh`, `sigmoid`, etc.) => Xavier
* ReLU-shaped activation (`relu`, `gelu`, `silu`, etc.) => Kaiming

For networks without Layer or Batch Normalization, that simple tweak will significantly increase convergance

### DIFF
--- a/src/nf/nf_dense_layer_submodule.f90
+++ b/src/nf/nf_dense_layer_submodule.f90
@@ -2,7 +2,7 @@ submodule(nf_dense_layer) nf_dense_layer_submodule
 
   use nf_activation, only: activation_function
   use nf_base_layer, only: base_layer
-  use nf_random, only: random_normal
+  use nf_random, only: random_normal, random_xavier, random_he
 
   implicit none
 
@@ -125,8 +125,18 @@ contains
     ! Weights are a 2-d array of shape previous layer size
     ! times this layer size.
     allocate(self % weights(self % input_size, self % output_size))
-    call random_normal(self % weights)
-    self % weights = self % weights / self % input_size
+    if (&
+        self % activation_name == 'relu' &
+        .or. self % activation_name == 'leaky_relu' &
+        .or. self % activation_name == 'celu' &
+    ) then
+      call random_he(self % weights, self % input_size)
+    elseif (self % activation_name == 'sigmoid' .or. self % activation_name == 'tanhf') then
+      call random_xavier(self % weights, self % input_size)
+    else
+      call random_normal(self % weights)
+      self % weights = self % weights / self % input_size
+    end if
 
     ! Broadcast weights to all other images, if any.
 #ifdef PARALLEL
--- a/src/nf/nf_random.f90
+++ b/src/nf/nf_random.f90
@@ -6,7 +6,7 @@ module nf_random
   implicit none
 
   private
-  public :: random_normal
+  public :: random_normal, random_he, random_xavier
 
   real, parameter :: pi = 4 * atan(1.d0)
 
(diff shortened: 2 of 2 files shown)


milancurcic:
Thanks, Michael, this is definitely needed.

About 1.5 years ago I started an Initializers PR (https://github.com/modern-fortran/neural-fortran/pull/151) but forgot about it. Basically it follows a similar pattern to how activations and optimizers are done in NF, which allows complete customization if specified, and sane defaults (like the ones you have here) if unspecified.

Do you think it would work well?

OneAdder:
@milancurcic Yes, I think #154 will work!

src/nf/nf_dense_layer_submodule.f90
```diff
@@ -125,8 +125,18 @@ module subroutine init(self, input_shape)
...
-    call random_normal(self % weights)
-    self % weights = self % weights / self % input_size
+    if (&
+        self % activation_name == 'relu' &
+        .or. self % activation_name == 'leaky_relu' &
+        .or. self % activation_name == 'celu' &
+    ) then
+      call random_he(self % weights, self % input_size)
+    elseif (self % activation_name == 'sigmoid' .or. self % activation_name == 'tanhf') then
+      call random_xavier(self % weights, self % input_size)
```
jvdp1, line 135: Should these be as default? Or should the user be able to choose for another pseudo-random generator?

src/nf/nf_random.f90
```diff
@@ -23,4 +23,22 @@ impure elemental subroutine random_normal(x)
     x = sqrt(- 2 * log(u(1))) * cos(2 * pi * u(2))
   end subroutine random_normal
 
+  impure elemental subroutine random_he(x, n_prev)
+    !! Kaiming weight initialization
+    real, intent(in out) :: x
```
jvdp1, line 28:
```suggestion
    real, intent(out) :: x
```

src/nf/nf_random.f90
```diff
@@ -23,4 +23,22 @@ impure elemental subroutine random_normal(x)
...
+    integer, intent(in) :: n_prev
+    call random_number(x)
+    x = x * sqrt(2. / n_prev)
+  end subroutine random_he
+
+  impure elemental subroutine random_xavier(x, n_prev)
+    !! Kaiming weight initialization
+    real, intent(in out) :: x
```
jvdp1, line 36:
```suggestion
    real, intent(out) :: x
```

src/nf/nf_random.f90
```diff
@@ -23,4 +23,22 @@ impure elemental subroutine random_normal(x)
...
+    integer, intent(in) :: n_prev
+    real :: lower, upper
+    lower = -(1. / sqrt(real(n_prev)))
+    upper = 1. / sqrt(real(n_prev))
```
jvdp1, line 40:
```suggestion
    upper = 1. / sqrt(real(n_prev))
    lower = -upper
```

src/nf/nf_random.f90
```diff
@@ -23,4 +23,22 @@ impure elemental subroutine random_normal(x)
...
+    call random_number(x)
+    x = lower + x * (upper - lower)
```
jvdp1, line 42: Is this correct if `lower == -upper`?

src/nf/nf_dense_layer_submodule.f90
OneAdder: I like how it's done here: #151 
In the DL Framework of my dreams, I would have an option to pass the algorithm of weights initialization into a layer's constructor. So:
* **What I want**: #151 with Kaiming weights by default but it requires a lot of refactoring
* **Why I made this PR-draft**: it is correct from the mathematical standpoint, Xavier for S-shaped and He for .*elu. Will probably resolve https://github.com/modern-fortran/neural-fortran/issues/145 if added to Conv layer
* **How Torch does it**: Kaiming for everything. Not ideal, but covers vast majority of cases

PR #205, opened 2025-02-20
Embedding Layer

OneAdder:
# Input Embeddings Lookup Table (Trainable
## Core
In Natural Language Processing input data is often encoded as indices of tokens in a vocabulary. Those indices are converted into vectors using weights which are trainable.
In theory, similar behaviour can be achieved by spreading input data by the desired size of vectors and then put through `input2d` and `linear2d` layers. However, it is very inefficient as we'll have to do `matmul` each time instead of simply getting an element by indices.
So, I created the layer that does it efficiently. It does not have a gradient as it is intended as an input layer. But it is trainable (`get_params`, `get_gradients` and `set_params`).

## Positional
Apart from this core functionality, I also added positional encoding (output vectors + positions assigned to sin/cos waves). This will be needed for transformers.

## Python Reference
[Here](https://github.com/OneAdder/neural-fortran-references/blob/main/embedding.py): `torch.nn.Embedding` and a custom function for positional encoding.

### DIFF
--- a/src/nf/nf_embedding_layer_submodule.f90
+++ b/src/nf/nf_embedding_layer_submodule.f90
@@ -34,6 +34,8 @@ contains
       index = input(i)
       if (index > size(self % weights, 1)) then
         index = 1
+      elseif (index == 0) then
+        index = 1
       end if
       self % output(i, :) = self % weights(index, :)
     end do
--- a/src/nf/nf_embedding_layer_submodule.f90
+++ b/src/nf/nf_embedding_layer_submodule.f90
@@ -88,7 +88,7 @@ contains
     real, pointer :: w_(:) => null()
 
     w_(1: product(shape(self % weights))) => self % weights
-    params = [w_]
+    params = w_
   end function get_params
 
   module function get_gradients(self) result(gradients)
@@ -97,7 +97,7 @@ contains
     real, pointer :: dw_(:) => null()
 
     dw_(1: product(shape(self % dw))) => self % dw
-    gradients = [dw_]
+    gradients = dw_
   end function get_gradients
 
   module subroutine set_params(self, params)
--- a/src/nf/nf_embedding_layer_submodule.f90
+++ b/src/nf/nf_embedding_layer_submodule.f90
@@ -17,7 +17,6 @@ contains
     self % sequence_length = input_shape(1)
 
     allocate(self % output(self % sequence_length, self % model_dimension))
-    allocate(self % gradient(self % sequence_length, self % vocab_size))
 
     allocate(self % weights(self % vocab_size, self % model_dimension))
     self % weights = 0.1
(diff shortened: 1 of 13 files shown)


src/nf/nf_embedding_layer_submodule.f90
```diff
@@ -0,0 +1,121 @@
...
+    num_params = self % vocab_size * self % model_dimension
+  end function get_num_params
+
+  module function get_params(self) result(params)
+    class(embedding_layer), intent(in), target :: self
+    real, allocatable :: params(:)
+    real, pointer :: w_(:) => null()
+
+    w_(1: product(shape(self % weights))) => self % weights
+    params = [w_]
```
jvdp1, line 91: Could it be:
```suggestion
    params = w_
```

src/nf/nf_embedding_layer_submodule.f90
```diff
@@ -0,0 +1,121 @@
...
+  end function get_params
+
+  module function get_gradients(self) result(gradients)
+    class(embedding_layer), intent(in), target :: self
+    real, allocatable :: gradients(:)
+    real, pointer :: dw_(:) => null()
+
+    dw_(1: product(shape(self % dw))) => self % dw
+    gradients = [dw_]
```
jvdp1, line 100:
```suggestion
    gradients = dw_
```
?

test/test_embedding_layer.f90
```diff
@@ -0,0 +1,83 @@
...
+  logical :: ok = .true.
+
+  call test_simple(ok)
+  call test_positional(ok)
+
+  if (ok) then
+    print '(a)', 'test_embedding_layer: All tests passed.'
+  else
+    write(stderr, '(a)') 'test_embedding_layer: One or more tests failed.'
+    stop 1
```
jvdp1, line 15:
```suggestion
    error stop 1
```

OneAdder:
@jvdp1 @milancurcic Ready for review

jvdp1 (commented):
LGTM. Below are some minor suggestions.

src/nf/nf_network_submodule.f90
```diff
@@ -207,8 +208,11 @@ module subroutine forward_1d(self, input)
     integer :: n
 
     ! Set the input array into the input layer
-    select type(input_layer => self % layers(1) % p); type is(input1d_layer)
-      call input_layer % set(input)
+    select type(input_layer => self % layers(1) % p)
+      type is(input1d_layer)
+        call input_layer % set(input)
+      type is(embedding_layer)
+        call input_layer % forward(nint(input))
```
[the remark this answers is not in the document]
OneAdder: Good idea. It'll be a lot more efficient for dictionaries of real-life sizes (e.g. 100k tokens). I added the generics

milancurcic:
Thanks for the nudge. Will try to finish review and merge either today or Monday. Thank you for all the hard work!

milancurcic:
@OneAdder Can you add the Embedding layer entry to the table of layers in the README? I assume that based on it being provided in `nf` that it's meant to be consumed by users directly, rather than only internally by other layers.

OneAdder:
@milancurcic Readme updated. Thank you for resolving conflicts!

milancurcic (approved):
Thank you!

PR #208, opened 2025-02-24
Fc2d layer

OneAdder:
## Fully-Connected Layer for 2D Shapes

Also known as MLP, FeedForward, etc. A common component of neural networks, including transformers. The idea is very simple: first linear transformation => activation => second linear transformation.

**This is the last piece of tranformer architecture.**
When #203, #205 and this one are merged. We can start adding transformer encoders and decoders.

## Problem

Softmax derivative here is incorrect. [This implementation](https://github.com/modern-fortran/neural-fortran/blob/ed8b3404264c0932b8c6d712f1dc29203f3d69e8/src/nf/nf_activation.f90#L391) is actually prime of logistic function which does not equivalent to softmax.
Derivative of softmax w.r.t. to each element in input requires computation of of Jacobian matrix:

$jacobian_{i, j} = \pmatrix{\frac{dsofmax_1}{dx_1} & ... & \frac{dsofmax_1}{dx_j} \cr ... & ... & ... \cr\frac{dsofmax_i}{dx_1} & ... & \frac{dsofmax_i}{dx_j} }$
$\frac{dsoftmax}{dx_i} = gradient \times jacobian$

Where:

* $\frac{dsoftmax_i}{x_j} = softmax(x_j) \cdot (\alpha - softmax(x_i))$ where $\alpha$ is $1$ for $i = j$, $0$ otherwise

* $x$ is the input sequence

Similar to my implementation for MultiHead Attention [here](https://github.com/modern-fortran/neural-fortran/blob/ed8b3404264c0932b8c6d712f1dc29203f3d69e8/src/nf/nf_multihead_attention_submodule.f90#L63).

## Possible Solutions
It is not easy to resolve as `activation_function` doesn't accept input, so:
* Do nothing, I added crutch that throws an error when `softmax` is passed as activation
* Make softmax a layer without parameters rather than an activation function, this will work
* Make a wrapper `activation_layer` that extends `base_layer` and accepts activation function

### DIFF
--- a/example/simple_2d_mlp.f90
+++ b/example/simple_2d_mlp.f90
@@ -1,0 +1,35 @@
+program simple
+  use nf, only: dense, fc2d, flatten, linear2d, input, network, sgd, relu, tanhf
+  implicit none
+  type(network) :: net
+  real, allocatable :: x(:, :), y(:)
+  integer, parameter :: num_iterations = 25
+  integer :: n
+
+  print '("Simple")'
+  print '(60("="))'
+
+  net = network([ &
+    input(4, 5), &
+    fc2d(3, 2, activation=relu()), &
+    flatten(), &
+    dense(4, activation=tanhf()) &
+  ])
+
+  call net % print_info()
+
+  allocate(x(4, 5))
+  call random_number(x)
+  y = [0.123456, 0.246802, 0.9, 0.001]
+
+  do n = 0, num_iterations
+
+    call net % forward(x)
+    call net % backward(y)
+    call net % update(optimizer=sgd(learning_rate=0.05))
+
+    if (mod(n, 5) == 0) print *, n, net % predict(x)
+
+  end do
+
+end program simple
--- a/src/nf.f90
+++ b/src/nf.f90
@@ -11,7 +11,8 @@ module nf
     linear2d, &
     maxpool2d, &
     reshape, &
-    self_attention
+    self_attention, &
+    fc2d
   use nf_loss, only: mse, quadratic
   use nf_metrics, only: corr, maxabs
   use nf_network, only: network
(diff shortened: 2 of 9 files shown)


jvdp1:
@OneAdder Please forgive my ignorance here. Could you please clarify the distinction between the `fc2d` layer and the `dense` layer?

milancurcic:
Thanks @OneAdder for starting this. From your explanation I understand what this does.

Rather than introducing a composition of multiple operations as a single layer, I suggest that we build a basic building block first, and then if needed, we can add a "shallow-wrapper" layer around those elementary layers.

Specifically, rather than introducing here a new layer that does "first linear transformation => activation => second linear transformation", I suggest we simply introduce a `dense2d` layer which is the same as `dense` but that works on 2-d inputs.

Then, the operation proposed here would be: `dense2d(activation) => dense2d(linear)`. What do you think?

And thanks for pointing out the incorrect softmax derivative. I don't even recall how and why I did that.

OneAdder:
@milancurcic It makes sense. I can do it. Should we merge this and then refactor it or the other way around?
BTW, I think we should actually make a consistent API for combined layers. Something along the lines of the following: `base_layer` is inherited by `combined_layer` which implements get and set params and gradients methods which point to the params of the layers that make up the combined layer. Combined layers extend `combined_layer` class

milancurcic:
Thanks, @OneAdder. If you agree, I suggest that here we simply provide a 2-d version of an existing `dense` layer (I suggest `dense_2d`) which accepts an activation function as well as a `linear` activation as a special case.

Good ideas for `combined_layer` but let's discuss it in a separate issue. I opened #217.

milancurcic:
Actually, since we already have `linear2d`, should we just refactor it to accept an activation function and thus call it `dense2d`? Then, creating `dense2d(..., activation="linear")` would give us `linear2d`.

OneAdder:
@milancurcic on it

PR #209, opened 2025-02-25
Multihead Attention Fixes

OneAdder:
### Minor Changes to MHA
1. Memory allocation optimization
2. Making attention masks usable (with test)
3. Minor cleanup (unused variables etc). BTW, perhaps, add some static analyzer to CI?
4. A bit of reshuffling inside MHA forward and backward calls for better inheritance. Reason for those changes is that I'm making Llama Attention which is subclass of MHA. Source here: https://github.com/OneAdder/llm.f/blob/master/src/llama_attention.f90

### DIFF
--- a/src/nf/nf_multihead_attention.f90
+++ b/src/nf/nf_multihead_attention.f90
@@ -41,17 +41,17 @@ module nf_multihead_attention_layer
     real, allocatable :: o_input(:, :)
 
     ! temporary storages for forward and backward passes
-    real, allocatable, private :: normalized_attention(:, :, :)
-    real, allocatable, private :: q_or_dq(:, :, :)
-    real, allocatable, private :: k_or_dk(:, :, :)
-    real, allocatable, private :: v_or_dv(:, :, :)
-    real, allocatable, private :: d_output(:, :, :)
-    real, allocatable, private :: v_heads(:, :, :)
-    real, allocatable, private :: k_heads(:, :, :)
-    real, allocatable, private :: q_heads(:, :, :)
-    real, allocatable, private :: d_sdpa(:, :)
-    real, allocatable, private :: jacobian(:, :)
-    real, allocatable, private :: d_normalize(:, :, :)
+    real, allocatable :: normalized_attention(:, :, :)
+    real, allocatable :: q_or_dq(:, :, :)
+    real, allocatable :: k_or_dk(:, :, :)
+    real, allocatable :: v_or_dv(:, :, :)
+    real, allocatable :: d_output(:, :, :)
+    real, allocatable :: v_heads(:, :, :)
+    real, allocatable :: k_heads(:, :, :)
+    real, allocatable :: q_heads(:, :, :)
+    real, allocatable :: d_sdpa(:, :)
+    real, allocatable :: jacobian(:, :)
+    real, allocatable :: d_normalize(:, :, :)
   contains
 
     procedure :: common_backward
--- a/src/nf/nf_multihead_attention.f90
+++ b/src/nf/nf_multihead_attention.f90
@@ -56,6 +56,8 @@ module nf_multihead_attention_layer
 
     procedure :: common_backward
     procedure :: common_forward
+    procedure :: sdpa_forward
+    procedure :: sdpa_backward
     procedure :: get_num_params
     procedure :: get_params
     procedure :: get_gradients
@@ -101,6 +103,17 @@ module nf_multihead_attention_layer
       real, intent(in) :: query(:, :), key(:, :), value(:, :)
       real, optional, intent(in) :: attention_mask(:, :)
     end subroutine common_forward
+
+    pure module subroutine sdpa_forward(self, attention_mask)
+      class(multihead_attention_layer), intent(in out) :: self
+      real, intent(in), optional :: attention_mask(:, :)
+    end subroutine sdpa_forward
+
+    pure module subroutine sdpa_backward(self, gradient, attention_mask)
+      class(multihead_attention_layer), intent(in out) :: self
+      real, intent(in) :: gradient(:, :)
+      real, intent(in), optional :: attention_mask(:, :)
+    end subroutine sdpa_backward
 
     pure module subroutine init(self, input_shape)
       !! Initialize the layer data structures.
--- a/src/nf/nf_multihead_attention.f90
+++ b/src/nf/nf_multihead_attention.f90
@@ -158,18 +158,18 @@ module nf_multihead_attention_layer
     end function get_num_params
 
     module function get_params(self) result(params)
-      class(multihead_attention_layer), intent(in), target :: self
+      class(multihead_attention_layer), intent(in) :: self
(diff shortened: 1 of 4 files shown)


OneAdder:
I figured out that a little refactoring would benefit greatly. I'll add it and then this will be ready for review

src/nf/nf_multihead_attention.f90
```diff
@@ -68,25 +83,38 @@ end function multihead_attention_layer_cons
...
       !! General forward propagation for MultiHead Attention Mechanism
       !! Might be used for both Self and Cross Attention
       !! Self Attention: pass the same value thrice
       !! Cross Attention: pass three values for your query, key and value
       class(multihead_attention_layer), intent(in out) :: self
       real, intent(in) :: query(:, :), key(:, :), value(:, :)
+      real, optional, intent(in) :: attention_mask(:, :)
     end subroutine common_forward
 
+    pure module subroutine sdpa_forward(self, attention_mask)
```
OneAdder, line 107: Put Scaled Dot Product Attention into a separate method. This adds more flexibility.
In some cases we need to do manipulations with input projections, such as _KV Caching_ for LLama and Qwen2.

OneAdder:
Ready for review

jvdp1 (commented):
LGTM!

src/nf/nf_multihead_attention.f90
```diff
@@ -39,10 +39,25 @@ module nf_multihead_attention_layer
...
+    real, allocatable :: k_heads(:, :, :)
+    real, allocatable :: q_heads(:, :, :)
+    real, allocatable :: d_sdpa(:, :)
+    real, allocatable :: jacobian(:, :)
+    real, allocatable :: d_normalize(:, :, :)
   contains
 
     procedure :: common_backward
     procedure :: common_forward
+    procedure :: sdpa_forward
```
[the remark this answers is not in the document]
OneAdder: will do!

src/nf/nf_multihead_attention_submodule.f90
```diff
@@ -14,51 +13,91 @@ module function multihead_attention_layer_cons(n_heads) result(res)
...
-    q_heads = self % split_heads(self % query_layer % output)
+    self % d_output = self % split_heads(self % output_layer % gradient)
 
     ! iterate over heads to calculate deltas for each of them
     do concurrent(head = 1: self % n_heads)
-      dv(:, :, head) = matmul(transpose(self % attention_matrix(:, :, head)), d_output(:, :, head))
+      self % v_or_dv(:, :, head) = matmul(&
+          transpose(self % attention_matrix(:, :, head)),&
+          self % d_output(:, :, head)&
+      )
```
milancurcic, line 93: Since this PR is doing some optimizations, there's an opportunity here as well. Intrinsic `matmul` can often be slower than writing out the matrix multiplication by hand using loops, +, and -. `transpose` creates a temporary copy, and can be avoided if we're looping and assigning values explicitly. When @jvdp1 made this optimization in `dense % forward`, I think it cut the compute time in half.

src/nf/nf_multihead_attention_submodule.f90
```diff
@@ -14,51 +13,91 @@ module function multihead_attention_layer_cons(n_heads) result(res)
...
 
       ! calculate delta for attention matrix
-      d_sdpa = matmul(d_output(:, :, head), transpose(v_heads(:, :, head)))
+      self % d_sdpa = matmul(self % d_output(:, :, head), transpose(self % v_heads(:, :, head)))
```
milancurcic, line 96: As above, similar could be done here.

src/nf/nf_multihead_attention_submodule.f90
```diff
@@ -82,79 +121,19 @@ pure module subroutine common_backward(self, input, gradient)
         ! multiply output of softmax by temp jacobian matrix
         ! For computational efficiency (avoid more temp storages), scaling is also done here
         ! reshapes: [3] -> [1, 3] @ [3, 3] = [1, 3] -> [3]
-        d_normalize(seq, :, head) = reshape(matmul(&
-            reshape(d_sdpa(seq, :), [1, self % sequence_length]),&
-            jacobian * self % scaling_factor&
+        self % d_normalize(seq, :, head) = reshape(matmul(&
+            reshape(self % d_sdpa(seq, :), [1, self % sequence_length]),&
+            self % jacobian * self % scaling_factor&
```
milancurcic, line 126: As above, in this case it may be even worse since we're doing `reshape(matmul(reshape, ...))`

src/nf/nf_multihead_attention_submodule.f90
```diff
@@ -82,79 +121,19 @@ pure module subroutine common_backward(self, input, gradient)
...
         ), [self % sequence_length])
       end do
 
       ! calculate delta for query
-      dq(:, :, head) = matmul(d_normalize(:, :, head), k_heads(:, :, head))
+      self % q_or_dq(:, :, head) = matmul(self % d_normalize(:, :, head), self % k_heads(:, :, head))
 
       ! calculate delta for key, attention matrix should be transposed unlike for query
-      dk(:, :, head) = matmul(transpose(d_normalize(:, :, head)), q_heads(:, :, head))
+      self % k_or_dk(:, :, head) = matmul(transpose(self % d_normalize(:, :, head)), self % q_heads(:, :, head))
```
milancurcic, line 134: As above

milancurcic (approved):
Thanks @OneAdder this looks good, see a few comments for further optimization that we've done before in `dense % forward`. Worth trying if you don't mind, or let me know if I can help.

milancurcic:
> Minor cleanup (unused variables etc). BTW, perhaps, add some static analyzer to CI?

I don't know how to do this and I'm too lazy to learn since I have this in my editor already :)

OneAdder:
@milancurcic @jvdp1 Do you think replacing matmuls with custom `do concurrent` is a good idea? The issue I see here is that we are stripping the power of `-fexternal-blas` which is super fast. Perhaps, the pointers approach could be a better idea? What I was thinking: add `view` function that returns a pointer and works as `reshape` but without an extra copy. Similar to [how PyTorch does it](https://pytorch.org/docs/stable/generated/torch.Tensor.view.html)

milancurcic:
Thank you for all the work so far, @OneAdder, and good luck with the job search! Feel free to get back to this whenever you're ready and I'll resume with the reviews of other PRs.

milancurcic:
> @milancurcic @jvdp1 Do you think replacing matmuls with custom do concurrent is a good idea? The issue I see here is that we are stripping the power of -fexternal-blas which is super fast. Perhaps, the pointers approach could be a better idea? What I was thinking: add view function that returns a pointer and works as reshape but without an extra copy.

Indeed, it's more complicated, as you write. Since the way you have it now is correct, any performance optimization can be left for a separate PR and we can investigate in more detail what approach we should take. So I'll just go ahead an merge this PR and we can revisit the optimizations later.

PR #212, opened 2025-02-28
linear2d_layer_no_bias: add optional argument to disable biases

OneAdder:
### Add option to disable biases
Simple additional option. Removing biases helps speed up inference. This technique is used in Llama and, to an extent, in Qwen. Being able to disable biases here is a simple quality of life addition

### DIFF
--- a/src/nf/nf_layer_constructors.f90
+++ b/src/nf/nf_layer_constructors.f90
@@ -213,12 +213,14 @@ module nf_layer_constructors
         !! Resulting layer instance
     end function reshape
 
-    module function linear2d(out_features) result(res)
+    module function linear2d(out_features, biases) result(res)
       !! Rank-2 (sequence_length, out_features) linear layer constructor.
       !! sequence_length is determined at layer initialization, based on the
       !! output shape of the previous layer.
       integer, intent(in) :: out_features
         !! Number of output features
+      logical, optional :: biases
+        !! Whether to use biases or not
       type(layer) :: res
         !! Resulting layer instance
     end function linear2d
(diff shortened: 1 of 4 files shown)


jvdp1:
> Being able to disable biases here is a simple quality of life addition

Should we extend this to all types of layers?

OneAdder:
@jvdp1 yep, I think we should

OneAdder:
Perhaps, I should rename it to FeedForward or MLP. Thoughts?

milancurcic:
I don't think MLP is appropriate because it refers to a network. We already have `dense`. Since you mention naming, and I don't recall if we discussed this in the original `linear2d` PR: `linear` layer is just a special case of a `dense` layer without an activation (which we have via the `linear` activation function). Should this be called

milancurcic:
This also means that we could have a generic `maxpool` after all. We could make it so that `maxpool1d` is invoked by `maxpool(pool_size=2)`, while `maxpool2d` is invoked by `maxpool(pool_size=[2, 2])`, because a scalar and array are type-kind-rank distinguishable. However, using this pattern we couldn't extend this to `maxpool3d` because `maxpool(pool_size=[2, 2, 2])` would not be distinguishable from the 2d variant.

Alternatively, we could do:

```fortran
maxpool(pool_width=2)  ! maxpool1d
maxpool(pool_width=2, pool_height=2)  ! maxpool2d
maxpool(pool_width=2, pool_height=2, pool_depth=2)  ! maxpool3d
```

Same approach could apply to `conv` layers.

The more I think about this the more I like it. What do you think?

OneAdder:
@milancurcic Disregard the naming comment. It belongs in a different PR: #208 
I like the maxpool ideas

OneAdder:
@milancurcic Should we merge it? I really would like to do it before the dense2d refactoring

milancurcic:
Yes, that's fine, please go ahead and squash-merge, thank you!

PR #214, opened 2025-03-14
Poolings

ricor07:
Starting to create PR. Still a stage 0 draft

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,8 +38,12 @@ add_library(neural-fortran
   src/nf/nf_layer_constructors_submodule.f90
   src/nf/nf_layer.f90
   src/nf/nf_layer_submodule.f90
+  src/nf/nf_locally_connected_1d_submodule.f90
+  src/nf/nf_locally_connected_1d.f90
   src/nf/nf_loss.f90
   src/nf/nf_loss_submodule.f90
+  src/nf/nf_maxpool1d_layer.f90
+  src/nf/nf_maxpool1d_layer_submodule.f90
   src/nf/nf_maxpool2d_layer.f90
   src/nf/nf_maxpool2d_layer_submodule.f90
   src/nf/nf_metrics.f90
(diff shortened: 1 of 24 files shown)


ricor07:
@milancurcic just a brief question: if maxpool2d requires input3d this means that I can't write a maxpool3d without an input4d? (Which of course I'm not going to implement)

milancurcic:
I don't know. What input rank does maxpool3d need and what does it output?

milancurcic:
I think that's correct. For maxpool2d naming I merely followed an existing convention (in Keras I think) where 2d refers to pooling over 2d, with an extra dimension for channels. If we follow the same for maxpool3d, it would be for 4d total, as you write. However I'm not familiar with applications that pool over 3d, so I leave it to you to determine if this is a common/needed use case.

ricor07:
I may be wrong, but I fear that in neural-fortran you called maxpool1d what in tensorflow is called maxpool2d (https://www.tensorflow.org/api_docs/python/tf/keras/layers/MaxPool2D) and neural-fortran's maxpool2d is tensorflow's maxpool3d. In this way, we have no tensorflow's maxpool1d. Can you please have a look? I'm not sure of this

milancurcic:
I think I understand the confusion now. The naming is consistent with Keras. What makes it seem inconsistent is that in neural-fortran `maxpool2d` I never implemented setting different pool sizes in x- and y-dimensions, and `pool_size` is an integer, so you can only do square-shaped pools.

From the above linked docs:

> Downsamples the input along its spatial dimensions (height and width) by taking the maximum value over an input window (of size defined by pool_size) for each channel of the input. 

This means that Keras's `MaxPool2D` takes 3D input and produces 3D output. Neural-fortran's `maxpool2d` does the same.

ricor07:
Hello. Any updates from your side? Sorry for not having given progresses but these days i'm very busy.

milancurcic:
Thanks @ricor07 for the effort so far and no rush. Whenever you have time. I've been busy with travel past week or so but am otherwise focused on reviewing other open PRs. Is there something specific about this one that you'd like me to look at? In general I don't review PRs until they're marked ready for review (i.e. not a draft).

PR #227, opened 2025-08-03
get_set removal

Riccardo231:
Starting to work on the code; still a draft, maybe have a look so to tell me if somebody is visibly wrong. 

(I thought i was working on my fork but instead i created a branch on the main project, i hope this is not a problem. Thank you)

### DIFF
--- a/src/nf/nf_conv1d_layer.f90
+++ b/src/nf/nf_conv1d_layer.f90
@@ -35,7 +35,6 @@ module nf_conv1d_layer
       procedure :: get_num_params
       procedure :: get_params_ptr
       procedure :: init
-      procedure :: set_params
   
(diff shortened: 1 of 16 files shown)


Riccardo231:
Hello, I've also removed set_params()

What I wanted to discuss with you is that in the layer embedding_layer there are only weights and not biases, so I cannot properly use the get_params_ptr, I think you already noticed because you didn't implement it.

What should I do? I think that adding biases to the layer wouldn't break anything.

Plus, in self_attention_layer I cannot find the submodule and the get_set procedures. Where are they?

When I'll be done with these two layers I can finally remove every instance of the getset procedures. Thanks, I'll wait for an answer

PR #230, opened 2025-09-05
Fixing segmentation error on test_flatten_layer

Riccardo231:
Made AI rewrite the code asking it to avoid possible segerrors and apparently it worked. Have a look .

### DIFF
--- a/test/test_flatten_layer.f90
+++ b/test/test_flatten_layer.f90
@@ -1,126 +1,171 @@
 program test_flatten_layer
 
   use iso_fortran_env, only: stderr => error_unit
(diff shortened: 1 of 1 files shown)


milancurcic (changes requested):
Sorry, I can't approve this because there is no explanation of what was wrong and what was fixed. A lot of code that works was also re-written in different style, which is not welcome. I think a good use would be for the AI to identify the cause of the segfault, and for only that part of the code to be changed.

jvdp1:
I think that the  issue is deeper than that, because I get a Segfault with my own application too. I'll to get a closer look later.

PR #232, opened 2025-09-21
Addition of the functions get_optimizer_by_name and get_name for optimizer DT

jvdp1:
Addition of the functions 
* `get_optimizer_by_name`
* `get_name` for optimizer DT

I also set the function `get_activation_by_name` as `pure`

### DIFF
--- a/src/nf/nf_activation.f90
+++ b/src/nf/nf_activation.f90
@@ -733,7 +733,7 @@ contains
   end function eval_3d_celu_prime
 
   ! Utility Functions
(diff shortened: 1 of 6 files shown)


milancurcic:
For getting the optimizer name, what do you think about setting a `character(*), parameter :: name` as a member to each concrete optimizer type, so then the name could be `optimizer % name`.

jvdp1:
> What is the need for this? Is it for the user to be able to instantiate an optimizer by a string? 

Yes, it is. In my case, the user provides the name of the optimizer in a TOML file.

PR #235, opened 2025-09-28
Documentation: parallel builds with flang + Caffeine

rouson:
This PR includes the commit(s) from PR #234.  The current PR edits the README.md file to include instructions for parallel runs of neural-fortran with the LLVM `flang-new` compiler and the [Caffeine](https://go.lbl.gov/caffeine) parallel runtime library.

PR #236, opened 2025-09-29
Demo julienne

rouson:
This PR includes the commits from #235 and adds a demonstration of running the `linear_2d_layer` unit tests using the [Julienne](https://go.lbl.gov/julienne) correctness-checking framework.  The PR adds two files, the initial versions of which were automatically generated by Julienne's `scaffold` app
    
* test/driver.f90 - main program (unmodified `scaffold` output)
* test/linear_2d_layer_test_m.f90 (`scaffold` output edited to incorporate code from the pre-existing text)
    
The PR also edits the `fpm.toml` file to add Julienne 3.2.1 as a development dependency so that it is only downloaded and built if the tests are being built and run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,9 @@ jobs:
         fpm-version: "v0.10.1"
     - uses: actions/checkout@v4
     - name: Compile
-      run: fpm build --profile debug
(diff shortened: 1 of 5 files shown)


rouson:
Oh... and only image 1 prints results, which of course is important when testing with a large number of images.

PR #237, opened 2025-10-13
More concise testing

milancurcic:
Alternative to #236.

Thank you for demoing Julienne, Damian, and for the inspiration. I've thought about the minimal added code needed for NF to make more concise tests without an additional dependency. This PR shows it on two test suites, `test_dense_layer` and `test_dense_network`. 

In summary:

* Adds the `tuff` module in tuff.f90 (72 LOC), which provides `test` and `test_result`.
* `test` can take a logical expression or a user-provided function that is `test_result`-typed.
* `test` can act as a suite driver for a mix of logical-expr tests and user-function tests.
* That's it.

Obviously less powerful than Julienne but at the same time much simpler to use (for me).

If you notice any serious caveats, let me know. In my view, nothing more is needed for NF and I'd prefer a simpler approach than bringing in an external dependency.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,3 +1,5 @@
+add_library(tuff tuff.f90)
+
 foreach(execid
   input1d_layer
   input2d_layer
@@ -27,7 +29,7 @@ foreach(execid
   metrics
   )
   add_executable(test_${execid} test_${execid}.f90)
-  target_link_libraries(test_${execid} PRIVATE neural-fortran ${LIBS})
+  target_link_libraries(test_${execid} PRIVATE tuff neural-fortran ${LIBS})
 
   add_test(NAME test_${execid} COMMAND test_${execid})
 endforeach()


rouson:
@milancurcic I see now that the above output is the intended behavior for a failing test.  

Here are few more thoughts on what would likely save time for an. unfamiliar user or new developer in interpreting the test output:

I think the compiler output `stopping due to failed executions` could be confusing .  At first, `failed executions` suggested to me that a test crashed or somehow terminated early.  I therefore inserted my own stops in the program at various intermediate points before ultimately concluding that the program was actually running all the way to what is essentially its end.  It would be really useful to have a formatted summary, including pass/fail tallies, at the very end of the entire test-suite execution.   That's the motivation for having just one main program and putting the actual  tests in modules. I'm adding this as item 6 in the list that I wrote in my previous comment.

Also, I suggest indenting the text `incorrect updated weights.. failed` to clear that it's information that is grouped with one particular test.  I also have a slight preference for that information to come after the summary for that corresponding test (i.e., after `test_linear2d_layer: One or more tests failed.`).

milancurcic:
The rationale for `stop 1` at the end of a failing program is to set the non-zero exit code. How else would you do this without `stop`? Or do you suggest a failing suite should exit with 0?

milancurcic:
I'll merge this as a start, and other suites can be adapted in a separate PR.

PR #244, opened 2026-06-16
Addition of CIFAR10 and CIFAR100

Riccardo231:
Addressing #242, added datasets CIFAR10 and CIFAR100. 
Some things I didn't like and for which I ask for your review:

- couldn't download CIFAR with the already built-in read_binary_file. Maybe I could've tried harder, but I preferred to write another function. Hope it's ok.
- Sadly, CIFAR10 and CIFAR100 don't share the same structure. CIFAR10 consists in lines of 1 byte for the label and 3072(3x32x32) for each image. CIFAR100 has two labels, the first label ranging between 0 and 99 and the second ranging between 0 and 19 (the so-called "macroclasses"), and then the 3072 bytes. I didn't care about this second distinction, so I had to add a boolean value to the function(cifar_100) which skipped the second byte if true. I acknowledge the solution isn't elegant at all but I didn't know how to solve this.
- Accuracy is really low, just a little above 100% / n. of labels. I'm working from a bad PC so I can't wait hours for each iteration. If you have the possibility, please take at least a check by modifying something in the network so that accuracy gives at least a 30%. In this way we're sure about the quality of the data. 
- CIFAR10 consisted in 6 files with 10k images each. I had to introduce dummy variables. Let me know what do you think of the choice.


milancurcic:
Thanks! Looks great at a first glance, I'll review in more detail later today..

milancurcic:
I made this branch up to date with main and I added a test suite; there were a few small patches in the CIFAR loader that my Codex agent identified (around not checking for presence of optional args); this is addressed.

I also see there is an unrelated update to the `flatten()` test suite. I think it's fine if it stays here but I will re-write it to follow tuff.f90 approach.

Regarding only the first class for CIFAR-100, I think it's fine for now, and if a need comes up for subclasses we can fix it later.

I think we should separate `read_cifar` into `read_cifar10` and `read_cifar100`. Current approach with both in the same procedure is cumbersome.

milancurcic:
I haven't yet tried the CIFAR examples; I'll try to make them work

Riccardo231:
Ok so this weekend I'll be working on the CIFAR adjustments.
Relating to the flatten issue, I didn't even remember I committed a patch. If you can work on it, I'd be glad.

Riccardo231:
Have a look now. I have no idea why I failed the test_conv2d_network since on terminal it's all ok and I haven't touched anything of it.

milancurcic:
Yeah that test occasionally (rarely) fails to converge, I'll try to fix it at some point

Riccardo231:
Should I make other fixes?

Riccardo231:
Hello, did this pause?

milancurcic:
I'll look at the new example and tests and let you know.

